PR #19, opened 2016-04-21
distributed computation via maps interface.

petehellyer:
Hi,

I've been thinking a little bit about how to adapt this code to run across a large distributed compute cluster (particularly using scoop - http://scoop.readthedocs.org/en/0.7/). I stopped a little short of specifically re-writing to depend on scoop, but I did instead re-do a few bits and pieces to allow the user to define a custom map function through which the initialisation and optimisation routine are run instead of loops. The user can then pick whatever future of map they fancy (which in my case is scoop)

I also made a couple changes to the maximisation routine to make it a bit more flexible in the case of using a futurised map. For example, you might if you're running across a cluster, want to run a few evaluations in parallel at each iteration. I've allowed for that by adding an n_evals argument to maximise(). 

Anyhow, thought some of these changes might be useful in general.

BW

### DIFF
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -58,8 +58,7 @@ def acq_max(ac, gp, y_max, bounds):
 
 
 class BayesianOptimization(object):
-
-    def __init__(self, f, pbounds, verbose=1):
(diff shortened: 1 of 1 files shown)


fmfn:
This looks very interesting. I'll take a closer look over the the weekend, but I like the idea.

Thanks.

PR #21, opened 2016-06-02
Changed default correlation model to the Matern 5/2 kernel

avonmoll:
According to Snoek et al. the Matern 5/2 kernel performed the best in empirical
tests. Also, this commit changed the default acquisition function to 'ei' to
match the docstring.

### DIFF
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -56,6 +56,51 @@ def acq_max(ac, gp, y_max, bounds):
     # point technicalities this is not always the case.
     return np.clip(x_max, bounds[:, 0], bounds[:, 1])
 
+def matern52(theta, d):
+    """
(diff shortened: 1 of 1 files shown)


avonmoll:
P.S. I'm also separately attempting to implement sampling over Gaussian Process covariance parameters (as described in [Slice sampling covariance hyperparameters of latent Gaussian models](http://papers.nips.cc/paper/4114-slice-sampling-covariance-hyperparameters-of-latent-gaussian-models.pdf)) which was also recommended by Snoek et al. in Practical Bayesian Optimization of Machine Learning Algorithms.

My motivation is simply the fact that Snoek's own software for Bayesian Optimization, Spearmint, seems a bit overkill for my purposes (it requires installation of MongoDB, for instance).

fmfn:
LGTM

And I agree, the whole reason why I wrote this package was because I wanted something simpler than Spearmint.

Issue #28, opened 2016-10-25
Is there a way to cache results?


ternaus:
Is there a way to run optimization for N iterations, save results and next time continue from the same point?

Or maybe run optimization on different computers, combine results together and continue from this combined set of data points?

fmfn:
With the `initialize` method you can pass points in the parameter space for which you know the function's value and optimize from there.

The only caveat is that the API of initialize is not so great (see issue #25 for reference). It is an easy fix, I just haven't got around doing it. But if you read the method implementation you should be able to extend it by manipulating the object's internal parameters directly.

ternaus:
As I understand, calling after training `res` returns points that were used during optimization, but it does not return results that were obtained during initialization. Am I right?

PR #30, opened 2016-10-25
Added method points_to_csv that saves known data points to csv file

ternaus:
Added method that saves known datapoints as a csv file.

ex:

```
       target        alpha      colsample_bytree        gamma
        -1166.19102       7.0034                0.6849       8.3673
        -1142.71370       6.6186                0.7314       3.5455
        -1138.68293       6.0798                0.9540       2.3281
        -1146.65974       2.4566                0.9290       0.3456
        -1160.32854       1.9821                0.5298       8.7863

```

Format is consistent with `initialize_df` method and as a result allows to address https://github.com/fmfn/BayesianOptimization/issues/28

### DIFF
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -337,7 +337,7 @@ class BayesianOptimization(object):
 
         :return: None
         """
-        import pandas as pd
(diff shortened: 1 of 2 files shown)


fmfn:
Thanks for this PR.

I think the pandas dependency could be avoided here, though, and instead use python's CSV module.

PR #34, opened 2016-11-27
add scaler

jonathanng:
for meta parameters that are of very different scales, this is very useful.

### DIFF
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -59,7 +59,7 @@ class BayesianOptimization(object):
         self.i = 0
 
         # Internal GP regressor
-        self.gp = GaussianProcessRegressor(kernel=Matern(), n_restarts_optimizer=25, normalize_y=True)
+        self.gp = GaussianProcessRegressor(kernel=Matern(), n_restarts_optimizer=25)
 
         # Utility Function placeholder
         self.util = None


fmfn:
The maximize method allows the user to pass parameters to the GP, I'm not sure this would work with how its implemented in this PR.

On this note, the one issue I have is that this takes some control away from the user. It is harder to engineer a Kernel in there's a scaling transformation you don't know about. On the other hand, it makes it easier to just "plug and play" and get some good results. Perhaps there's a mid ground?

jonathanng:
Yes, you make some good points.

I added a scaler option to maximize. Let me know what you think?

Erotemic:
FYI I fixed the merge conflicts in this branch, added tests, and added a demo demonstrating that using scaler=True returns a better result than scaler=False in at least one scenario. 

https://github.com/Erotemic/BayesianOptimization/tree/scalers

@jonathanng  I could either target the PR to your branch (which in turn would update this), or we could close this PR and I could simply target a PR at this repo. (Note, I fixed the merge conflicts by checking out your branch directly and rebasing, so your username is still tied to your commits)

jonathanng:
Whatever you think is better! Thanks!

fmfn:
@Erotemic I think targeting the PR to his branch and updating this one might be better, so we have the discussion in one place.

fmfn:
Closing since it's outdated and I'm not fully convinced this is needed as a feature.

PR #47, opened 2017-03-12
Fixes bug in initialize

### LINKED ISSUE #46: Bug in initialize function
`initialize` function assumes keys are sorted and appear after 'target' in lexicographic order.

Code to reproduce:

```python


def target(a):
    """ Function to return square of argument """
    return a**2


def target2(x):
    """ Same function with different name of argument """
    return x**2


def test_bo_initialization(bo, func):
    print('X: ', bo.x_init)
    print('Y: ', bo.y_init)
    mismatch = False
    for x, y in zip(bo.x_init, bo.y_init):
        if func(*x) != y:
            mismatch = True
    if mismatch:    
        print('ERROR: x_init and y_init do not match!')
    else:
        print('INFO: x_init and y_init match.')

        
bo = BayesianOptimization(target, {'a': (0, 10)})
bo.initialize(
    {
        'target': [1, 4, 9],
        'a': [1, 2, 3],
    })
test_bo_initialization(bo, target)
# mismatch is True!
# Output: 
# X:  [[1], [4], [9]]
# Y:  [1, 2, 3]
# ERROR: x_init and y_init do not match!

bo2 = BayesianOptimization(target2, {'x': (0, 10)})
bo2.initialize(
    {
        'target': [1, 4, 9],
        'x': [1, 2, 3]
    })
test_bo_initialization(bo2, target2)
# mismatch is False.
# Output: 
# X:  [[1], [2], [3]]
# Y:  [1, 4, 9]
# INFO: x_init and y_init match.
```

amanbh:
- Fixes #46 
- Adds example for initialize function

### DIFF
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -151,11 +151,26 @@ class BayesianOptimization(object):
         Method to introduce points for which the target function value is known
 
         :param points_dict:
+            dictionary with self.keys and 'target' as keys, and list of
+            corresponding values as values.
+
+        ex:
+            {
+                'target': [-1166.19102, -1142.71370, -1138.68293],
+                'alpha': [7.0034, 6.6186, 6.0798],
+                'colsample_bytree': [0.6849, 0.7314, 0.9540],
+                'gamma': [8.3673, 3.5455, 2.3281],
+            }
+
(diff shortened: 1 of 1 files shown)


fmfn:
Nice catch, thanks for the fix!

PR #67, opened 2017-10-11
Update X and Y during init

ajnisbet:
I often plot/monitor the optimisation each iteration in case I want to stop early or change some parameters. I do this by adding monitoring code to the target function.

Currently, the x and y values checked during `explore` and `init` are not accessible (stored on the `bo` object) until after both steps are complete, which can take a long time for expensive functions. 

This PR adds updates `self.X` and `self.Y` each iteration of the explore/init step. Most of the logic is converting `append` and `+=` to numpy equivalents. Adds a bit of complexity but doesn't change any other part of the API.

You could go all out and also update `bo.go.fit()` and `bo.res` each iteration, and add callback parameter 
 to be called after each iteration to avoid hacking the target function like xgboost and some sklearn functions. But just updating `X` and `Y` is the minimum info needed for a user to just do that themselves.

### DIFF
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -105,7 +105,7 @@ class BayesianOptimization(object):
             self.Y = np.append(self.Y, self.f(**dict(zip(self.keys, x))))
 
             if self.verbose:
-                self.plog.print_step(x, y_init[-1])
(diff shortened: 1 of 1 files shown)


fmfn:
Looks good. Did you try running the `usage.py` script?

ajnisbet:
Yup, same logs and solution.

PR #77, opened 2017-12-09
Fixes for Travis

Erotemic:
There were a few errors in my Travis scripts, which this PR fixes. 

I also added travis shields to the readme (if you enable codecov the repo should be setup for that as well)

Lastly, the travis tests were running pretty slow, so I added a hidden option `self._acqkw`, which allows you to control how many warmup / L-BFGS runs you do. This lets the tests run much quicker, and might be a useful hidden feature for a user.


fmfn:
👍

PR #84, opened 2018-01-30
Update helpers.py

zhangchuangnankai:
May be fail to minimize,the minimize will return res.fun with a array(0.0), and get res.fun will get an Exception

### DIFF
--- a/bayes_opt/helpers.py
+++ b/bayes_opt/helpers.py
@@ -60,7 +60,7 @@ def acq_max(ac, gp, y_max, bounds, random_state, n_warmup=100000, n_iter=250):
                        method="L-BFGS-B")
 
         # See if success
-        if res.success==False:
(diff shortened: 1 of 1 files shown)


bayes_opt/helpers.py
```diff
@@ -59,6 +59,10 @@ def acq_max(ac, gp, y_max, bounds, random_state, n_warmup=100000, n_iter=250):
                        bounds=bounds,
                        method="L-BFGS-B")
 
+        # See if success
+        if res.success==False:
```
fmfn: Can you change to `if not res.success:`? This is more pythonic.

zhangchuangnankai: OK, I have changed

fmfn:
Thanks for the contribution!

Issue #94, opened 2018-05-10
Support dependencies between optimization parameters


Engineero:
It would be great if we could enter dependencies between parameters for optimization. For example, if I have the parameter bounds:

``` python
pbounds = {'a': (0, 5),
           'b': (0, 10)}
```

and know that for my application, no meaningful results will come from parameter choices where `b < a`, I would like to be able to force the optimizer to only choose samples such that `b >= a`.

See [[M. T. M. Emmerich *et. al.* 2008]][0] for some discussion of this. That's just about the only paper that tries it that I can find so far.

[0]: http://repository.ubn.ru.nl/bitstream/handle/2066/72676/72676.pdf

Engineero:
See pull request #95

fmfn:
I'll take a look at the PR. But one idea that sometimes can be used is to parametrize your parameters differently, making the dependency more explicit:

instead of ```pbounds = {'a': (0, 5), 'b': (0, 10}``` with the condition `b < a`, you could change coordinates to: `x = a` and `y = b - a`, and define `pbounds = {'x': (0, 5), 'y': (0, 5)}`, and call function like so: `f(a=x, b=x + y)`.

meyermj:
I have a similar requirement, but different in a key way:
```
pbounds = {'a': (0, 1),
           'b': (0, 1),
           'c': (0, 1)}
```
where `a+b+c=1`. I could be overlooking something, but I don't think there's a way to change my parameterization/function call to enforce this condition. Any pointers would be greatly appreciated!

dkhachatrian:
@meyermj I think for your setup, you can solve it by recasting in terms of spherical coordinates.

First, since $a,b,c > 0$, there exists a one-to-one correspondence $x = \sqrt{a}, y = \sqrt{b}, z = \sqrt{c}$ so you can rewrite as $x^2 + y^2 + z^2 = 1$.

Now, convert to spherical coordinates, via something like:

```python
from math import atan2, sin, cos

#...

r = (x^2 + y^2 + z^2)**(1/2)
theta = atan2(y,x)
phi = atan2(z,(x^2 + y^2)**(1/2))
```

Now in spherical coordinates, your `pbounds` can be written as

```python
pbounds = {'r': (0,1), 'theta': (0, math.pi/2), 'phi': (0, math.pi/2)}
```

and you can convert back to $x,y,z$ via

```python
z = r*cos(phi)
y = r*sin(phi)*sin(theta)
[...]

PR #95, opened 2018-05-10
Feature/parameter dependency

Engineero:
Updated maximization routine to accept constraints using the SLSQP optimizer. Added tests. This resolves issue #94.

It looks like the system will still guess outside of the constraints sometimes, but the updated `sklearn_example`, at least in my testing, showed that it mostly respects the constraints. The example shown for the `maximize` method still finds the global max, despite the constraint trying to restrict it to values above the 5000th index.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@ scratch/
 .DS_Store
 bo_eg*.png
 gif/
+__pycache__
+.pytest_cache
 
 # Unit test / coverage reports
 htmlcov/
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -25,7 +25,7 @@ class BayesianOptimization(object):
             and SLSQP optimizers. Each dictionary has the fields:
             
             - type : str
-              Constraint type: ‘eq’ for equality, ‘ineq’ for inequality.
+              Constraint type: 'eq' for equality, 'ineq' for inequality.
             - fun : callable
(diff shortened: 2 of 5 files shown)


Engineero:
I just realized that pull request #45 addresses this same issue. I believe our approaches are a bit different, so it may be worth looking in to both to see what you like best. Please merge one of these soon!

fmfn:
Thank you @Engineero , I will take a look at both. I'm a bit behind reviewing issues and PRs but will get around doing that soon.

fmfn:
My fear with implementing constraints on the acquisition function optimization is that the algorithm might get stuck probing the same edge points. Since the Gaussian Process is not constrained it may yield optimal points outside the allowed bounds, leading the `acq_max` function to return the same edge point over and over and the GP not being updated (since no new knowledge is gained), in a never ending loop.

In my mind reparametrizing the target function (embedding the constraints in the new parametrization) seems like a more robust solution. However that can be hard or even impossible at times. Alternatively I could see a very clever choice or kernel also being a solution to this problem.

zhuyanhao:
Hello @fmfn / @Engineero ,

I'm new to this package and I'm using it to solve a constrained problem. I had a feeling that the acquisition function should also been changed as the original ac functions are no longer some type of uncertainty of the optimal solution. Is there any related workings since this post? 

Thanks,
Yanhao

bwheelz36:
Closing this as we have a constrain model implemented now in #344

PR #100, opened 2018-06-19
include callback, now we can save .csv each observation

rspadim:
from kaggle https://www.kaggle.com/c/home-credit-default-risk/discussion/59195#345409

include callback is important to save csv from each step, since xgb and others algoritms can crash with memory limits and lost work


fmfn:
We now have the concept of observables in the library, and with the next release (under way) I intend to include this functionality. Stay tuned.

PR #105, opened 2018-08-15
All observed points added to result dict

guitargeek:
Hello,

I feel all observed points should be added to the results dict, not only the ones observed after the Bayesian fit. The random points in the initialization and the points explored by the user are equally valid, right? It would be sad if the maximum is one of them and it's therefore missed :)

I thought the best way to ensure all observed points are added to the results dict is to just do that in `BayesianOptimization._observe_point()`.

Cheers,
Jonas

### DIFF
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -100,10 +100,16 @@ class BayesianOptimization(object):
         # Updates the flag
         self.initialized = True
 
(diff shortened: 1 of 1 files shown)


fmfn:
I'm working on this and a number of other fixes and improvements, so I'll close this PR since a good amount of refactor is taking place.

guitargeek:
Ok thank you very much for the info! Cheers

PR #107, opened 2018-09-07
Make API incremental {Gauging interest}

bwasti:
This isn't in a mergable state, but the idea here is to make the API incremental so that users don't have to cede control of the loop to this library.

This will make it easier to integrate into projects that already have optimization loops and generic exploration/optimization policies.

Any interest?

### DIFF
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -11,11 +11,8 @@ from .target_space import TargetSpace
 
 class BayesianOptimization(object):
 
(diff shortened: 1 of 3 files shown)


fmfn:
This has been on my TODO list for a while now. This change would also open the door for asynchronous optimization, or easily putting this library behind an API. So yeah, there's interest in that. I was planning on tackling it in October, we can sync up if you're still working on it.

fmfn:
I started working on this and expect to have it out in about a week or so. So I'll close this PR since I'm doing a good amount of refactoring to get this and other fixes/improvements done.

PR #108, opened 2018-09-19
* Adding fixed arguments

joswinkj:
Adding fixed arguments to the function call. Useful for machine learning tasks
where parameter optimization for different fixed arguments is useful

### DIFF
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -11,7 +11,7 @@ from .target_space import TargetSpace
 
 class BayesianOptimization(object):
 
-    def __init__(self, f, pbounds, random_state=None, verbose=1):
+    def __init__(self, f, pbounds, random_state=None, verbose=1 ,fixed_args={}):
         """
         :param f:
             Function to be maximized.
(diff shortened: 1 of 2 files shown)


fmfn:
Can you explain what is the purpose of fixed arguments? I'm not sure I get it.

Amir-Arsalan:
@fmfn I think this is exactly what I was looking for. Assume I have a score function that takes as input the arguments `X`, `Y` and `Z`. In my problem `X` is the variable for which I want to run Bayesian optimization to get its best value. However, other arguments/parameters such as `Y` and `Z` that are always fixed but needed to do the function evaluation. In practice, this is very useful when you have to initialize lots of BO instances in parallel.

I wonder if you can merge this pull request if things look good?

Amir-Arsalan:
@joswinkj Could you please if my explanation is correct?

fmfn:
@Amir-Arsalan So you have a function `f(x, y, z)` and you want to optimize for `x` while keeping `y` and `z` constant. Maybe you also want to run another process in parallel where you optimize for `y` and keep the other two constant. I understand the need to do this, but why not simply do something like:
```
bo = BayesianOptimization(f=lambda x: f(x=x, y=y_fixed, z=z_fixed), ...)
```
or even
```
from functools import partial
x_func = partial(f, y=y_fixed, z=z_fixed)
bo = BayesianOptimization(f=x_func, ...)
```
?

I'm sorry, but I'm not convinced yet that this feature is truly needed.

Amir-Arsalan:
@fmfn I tried compiling your code by locally merging this pull request using `setup.py` but for some reason the built package has whatever the master branch has here. Is there a way to build the package with this pull request?

PR #125, opened 2018-11-23
add parameters type description in list of parameters bounds

phypho:
Parameters type are included into pbounds: `pbounds = {'p1': [int,(0, 1)], 'p2': [float,(1, 100)]}`
After the minimization and to avoid floating result for integer parameter, the parameter returned corresponds to the minimum between the lower and higher integer values of the floating returned value.
This solution assumes that the acquisition to be minimized is regular. This should not works if the parameter is categorical.
I hope the explanations are clear enough. Anyway, the modifications in the code are not complicated.

### DIFF
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -65,7 +65,7 @@ class BayesianOptimization(Observable):
                            'max_params': None}
         self.res['all'] = {'values': [], 'params': []}
 
-        # non-public config for maximizing the aquisition function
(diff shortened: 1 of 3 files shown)


fmfn:
Ok, I could get behind these changes.

The only thing is, I'm working on a v1.0.0 release, and there's a good amount of internal changes. I would appreciate if you could make a pull request towards the v1.0.0 branch instead (or wait until it is released in the next week or so).

phypho:
OK, I am moving to the latest release

PR #131, opened 2018-11-29
Add parameter typing (only float and integer)

jmehault:
Add optional parameter "ptypes" in the BayesianOptimization class in order to indicate le type of each parameter to optimize into a dictionary. For now, every parameters should be typed. The goal is mainly to validate the method.
If a parameter type is declared as integer, then the extremal position is chosen as follow. The outcome (float) of the minimization function is transformed into two positions (lower and upper integer). The minimum is the minimal value took by the acquisition function among the lower and upper positions.
In the case of multiple integer parameters, the current implementation do not combine every lower and upper integer values. The only combinations tested are all lowers versus all uppers.

Wait for comments !

### DIFF
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -129,6 +129,7 @@ class BayesianOptimization(Observable):
             gp=self._gp,
             y_max=self._space.target.max(),
             bounds=self._space.bounds,
+            btypes=self._space.btypes,
             random_state=self._random_state
         )
 
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -64,13 +64,13 @@ class Observable(object):
 
 
 class BayesianOptimization(Observable):
-    def __init__(self, f, pbounds, random_state=None, verbose=2):
+    def __init__(self, f, pbounds, ptypes=None, random_state=None, verbose=2):
         """"""
         self._random_state = ensure_rng(random_state)
 
         # Data structure containing the function to be optimized, the bounds of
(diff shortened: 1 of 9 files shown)


Jie-Yuan:
author updated this?

phypho:
I am using the updates for my works. Some improvement might be done, but I am waiting for first comments about the solution I proposed. I don't know if it should be merged at the end.

AlbertAlonso:
May I suggest an alternative indication method. In order to make it so that indicating the type of one parameter should not mean doing a whole dictionary, something like this?
```
>>> pbounds = {'p1': (0,1), 'p2': (0, 100, int)}
>>> [item[1][:2] for item in pbounds.items()]
[(0, 1), (0, 100)]
>>> [float if len(item[1])<3 else item[1][2] for item in pbounds.items()]
[<class 'float'>, <class 'int'>]
```
where a 3 element to the tupple is added with the type, otherwise float is considered. 
Thoughts?

mjkvaak:
I very much like the added value for BO with typing of the parameters! However, this could be still improved so that the parameters would get returned casted in their corresponding type. I mean, even if I define, e.g. `max_depth:int`, the parameter will get fed to the functio-to-be-optimized  as `float`. This caused me some headache, when I was trying to figure out why my classifier kept on throwing fit-errors at me.

ChanderJindal (approved):
The code looks fine.

ChanderJindal:
Nice idea here, till now, in the 'black_box_function' I was converting that float to int, it isn't the best but happened to be a convenient solution.

till-m:
Closing in favour of #531.

PR #219, opened 2020-04-12
minor refactor expected improvement

ben-arnao:
Minor increase in computational efficiency by only computing the value once instead of twice


fmfn:
Thank you

PR #221, opened 2020-04-16
Kappa decay

ben-arnao:
Add a tunable parameter kappa decay that will allow you to make the optimization more exploitative over time. From a few tests I've run i get faster convergence to better maxima.


fmfn:
Thanks for the PR. I really like the idea of adding this. However I'm not a fan of recreating the `UtilityFunction` object like that, can you add a `update_kappa` (or similarly named) method to `UtilityFunction` and use that inside the maximize loop instead? That way this feature is also accessible outside the maximize function too.

bayes_opt/bayesian_optimization.py
```diff
@@ -162,9 +163,16 @@ def maximize(self,
         self._prime_queue(init_points)
         self.set_gp_params(**gp_params)
 
-        util = UtilityFunction(kind=acq, kappa=kappa, xi=xi)
+        def decay_kappa(kappa, iters, decay):
```
fmfn, line 166: Can we turn this into a `UtilityFunction` method?

bayes_opt/bayesian_optimization.py
```diff
@@ -162,9 +163,16 @@ def maximize(self,
...
+            for x in range(iters):
+                kappa *= decay
+            return kappa
+
         iteration = 0
         while not self._queue.empty or iteration < n_iter:
+            bo_iters = max(0, len(self.res) - init_points)
+            decayed_kappa = decay_kappa(init_kappa, bo_iters, kappa_decay)
+            util = UtilityFunction(kind=acq, kappa=decayed_kappa, xi=xi)
```
fmfn, line 175: Use the method to update the value here as opposed to creating a new object.

ben-arnao: I put the decay in UtilityFunction and also added an optional "delay" parameter to give users the option to delay kappa decay by X BO iters before kappa starts to decay.

I think this will be useful for ensuring a fixed amount of explorative BO in the beginning steps before decaying kappa to be more exploitative. As with original kappa decay it is optional and by default turned off so most users will not even know this is an option. Doesn't change the existing API either.

Lmk what you think.

bayes_opt/util.py
```diff
@@ -99,6 +99,9 @@ def utility(self, x, gp, y_max):
             return self._ei(x, gp, y_max, self.xi)
         if self.kind == 'poi':
             return self._poi(x, gp, y_max, self.xi)
+        
+    def decay_kappa(self, decay_factor):
```
fmfn, line 103: I meant passing the `decay_factor` (and `_delay` too) to the constructors, and then calling `util.update_params()` (or similar name) in the loop. All this logic `if kappa_decay != 1 and max(0, len(self.res) - init_points) - kappa_decay_delay > 0:` could reside inside the `UtilityFunction` object.

bayes_opt/bayesian_optimization.py
```diff
@@ -165,6 +167,10 @@ def maximize(self,
         util = UtilityFunction(kind=acq, kappa=kappa, xi=xi)
         iteration = 0
         while not self._queue.empty or iteration < n_iter:
+            # decay kappa by decay factor if BO iters exceeds kappa_decay_delay
+            if kappa_decay != 1 and max(0, len(self.res) - init_points) - kappa_decay_delay > 0:
```
fmfn, line 171: I rather not introduce this `if` statement here. It's at a different abstraction level than the surrounding code and it also only applies to one kind of utility function.

bayes_opt/util.py
ben-arnao: Ah i see. I will make the changes. For delay to be used however we will need to feed in the # of BO iters to the update function. Lmk how it looks now.

fmfn: `UtilityFunction` could keep an internal counter to keep track of delay.

ben-arnao: ok done

bayes_opt/bayesian_optimization.py
ben-arnao: is this still something you wanted to move forward with?

fmfn: Yes, absolutely. I've been a bit busy lately with work, so I'm slow to make comments and stuff. But we'll get this merged this week =)

fmfn:
The implementation is great! I'm ready to approve, but I have one small ask first. Would you mind rebasing and squashing all commits in a single one?

bwheelz36:
Hey guys, this is a really cool feature, but it took me a while to realise it was here - unless I missed it I don't think this is covered in the examples?
If it's helpful I could do a PR with an update to the advanced tour that demonstrated this feature?

PR #225, opened 2020-05-09
Add sequential domain reduction

osullivryan:
* Added a base `DomainTransformer` class for future transformers.
* Added a sequential domain reduction transformer which mutates the bounds of the problem (within their original ranges) to try and converge faster onto an optimal value. 
* Added a jupyter notebook which explains the parameters and gives an example with plotting. 
* Added a test function to test the new feature. 
* Fake8 some of the file. 

#### Background
For objective functions which can take quite a long time to solve, and especially in engineering applications, it is sometimes more worthwhile to converge on **an** optimal point quickly rather than try to find **the** optimal point. One of the ways to do this is by contracting the domain around the current optimal value as the search progresses. This PR implements a common algorithm for doing so.


osullivryan:
Sorry for the length! Most of that is in the jupyter notebook, though.

PR #226, opened 2020-05-09
Add sequential domain reduction

osullivryan:
* Added a base DomainTransformer class for future transformers.
* Added a sequential domain reduction transformer which mutates the bounds of the problem (within their original ranges) to try and converge faster onto an optimal value.
* Added a jupyter notebook which explains the parameters and gives an example with plotting.
* Added a test function to test the new feature.
* Fake8 some of the file.
#### Background
For objective functions which can take quite a long time to solve, and especially in engineering applications, it is sometimes more worthwhile to converge on an optimal point quickly rather than try to find the optimal point. One of the ways to do this is by contracting the domain around the current optimal value as the search progresses. This PR implements a common algorithm for doing so.

Sorry for the large PR! Let me know if you want me to break it up. Lucky most of this is in the jupyter notebook.


setup.py
```diff
@@ -2,7 +2,7 @@
 
 setup(
     name='bayesian-optimization',
-    version='1.1.0',
+    version='1.1.1',
```
fmfn, line 5: I say we bump this to 1.2.0 instead. Added functionality deserves it.

examples/domain_reduction.ipynb
```diff
@@ -0,0 +1,302 @@
+{
```
fmfn, line 1: Can we add a link to this notebook to the **Quick Start** list in the `README.md`?

fmfn:
Brilliant. Thank you!
I'll ship the new version sometime this weekend =)

PR #230, opened 2020-05-18
Added option to keep using an older log (Save -> Load -> Save)

### LINKED ISSUE #159: Save, load and save again
Hi,

Thanks for this useful library!

I wonder how to do `save - load - save - load` multiple times?
To achieve this, I use the code below:
```python
if os.path.exists(log_path):
    load_logs(optimizer, logs=[log_path])
    print('Load {} records from disk.'.format(len(optimizer.space)))

logger = JSONLogger(path=log_path)
optimizer.subscribe(Events.OPTMIZATION_STEP, logger)
```

But it seems that this line 
```
logger = JSONLogger(path=log_path)
```
will delete the orginal json file and previous records are lost.

Why do we need to do `os.remove(self._path)` in `JSONLogger.__init__()`?

AlbertAlonso:
Quick fix so that when loading a log file, the information it is not lost. Having an option to fix #159 easily. For backwards compatibility, reset is set to True.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,8 @@ optimizer.maximize(
 )
 ```
 
+By default the previous data in the json file is removed. If you want to keep working with the same logger, the `reset` paremeter in `JSONLogger` should be set to False.
+
 ### 4.2 Loading progress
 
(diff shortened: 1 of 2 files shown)


fmfn:
Thank you.

Issue #158, opened 2019-04-22
KeyError: 'Data point X is not unique'


PoCk3T:
Hi and congrats for this sweet piece of software, really love it :)

I've implemented an async solution very similar to the one in the example notebook, with 13 virtual machines connecting to a master server hosting the Tornado webserver (like in the notebook), and it seems like I'm now constantly running over the same symptoms on the master server (the server registering the tested points and their respective target in the async example):

```
Error: 'Data point [3.47653914e+04 2.10539196e+02 3.15656650e+00 6.77134492e+00  1.01962491e+01] is not unique'
Traceback (most recent call last):
  File "playground_master.py", line 72, in post
    self._bo.register(params=body["params"], target=body["target"])
  File "/usr/local/lib/python3.5/dist-packages/bayes_opt/bayesian_optimization.py", line 104, in register
    self._space.register(params, target)
  File "/usr/local/lib/python3.5/dist-packages/bayes_opt/target_space.py", line 161, in register
    raise KeyError('Data point {} is not unique'.format(x))
KeyError: 'Data point [3.47653914e+04 2.10539196e+02 3.15656650e+00 6.77134492e+00 1.01962491e+01] is not unique'
```

It seems like the BO is suggesting to the 13 VMs to test points (set of 5 values in my case) which are already tested, almost in a constant manner after ~ 800 points tested.

Here's my troubleshooting so far:
- Using the load/save examples in the front page of BO, I save all points tested in a JSON file and for this instance of the problem, I see 817 points already registered in the JSON
- It seems like this data point thrown in the traceback is indeed ALREADY in the JSON, with an associated target value
- The 13 slave VMs are still trying to ask for "suggestions" (further set of points to test), but it seems like BO gives set of points already tested ; I still see some rare cases when it's not tested yet and the counts case increase slightly to 818, 819 points.. (but most of the time the traceback is thrown)

I'm a little bit surprised you can end up with such scenario given my PBOUND is very broad, and so, has a lot to points to be worked on without having to test the same ones again:
```
            {'learning_timesteps': (5000, 40000),
            'timesteps_per_batch': (4, 72),
            'observation_loopback': (1, 20),
            'mmr__timeperiod': (7, 365),
            'model': (-0.49, 5.49) }
```

This is how I initialized the utility, which, as far as I understood, is responsible for suggesting points that were already tested:
`_uf = UtilityFunction(kind="ucb", kappa=2.576, xi=0)`

[...]

PR #232, opened 2020-05-23
Introduced use_only_unique_points parameter

LucaCappelletti94:
Hello and thank you for creating this library.

With this pull request, I wanted to target the issue #158, related to the repetition of search data points, and the impossibility to do transfer learning of Gaussian processes. 

I have added the necessary unit tests for the updates and every old test which was passing on master at this time is also passing within this pull request.


dajuguan:
hi, could u merge it to the pip registory?

PR #238, opened 2020-07-06
Added docstrings to the `main` functions.

AlbertAlonso:
I thought it may be useful to have some docstrings since it seems some of the issues are of people not knowing what some parameters do. 

I am up to go through all the code adding docstrings if it's fine for you. I have seen already some there so I've assumed the [numpy doc guide](https://numpydoc.readthedocs.io/en/latest/format.html#method-docstrings) was being followed. If not, my mistake.

Feel free to add/change any description. 

I couldn't find the use of `xi` anywhere, so I left it as [unused] so you could fix it.

### DIFF
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -70,15 +70,18 @@ class BayesianOptimization(Observable):
     Parameters
     ----------
     f: function
-        The function to optimize.
+        Function to be maximized.
+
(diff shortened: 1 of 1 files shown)


fmfn:
Thank you @AlbertAlonso !! This is excellent! If you're good writing more doc string it would be **very** much appreciated. I'll merge this for the time being so it can go out with the next release (which it's overdue).

PR #239, opened 2020-07-07
Queue

alonfnt:
Browsing the code I've seen a custom implementation of Queue. Since there is a queue on the Standard Library, why not use it?  it may be useful when doing threading? and there is less code to maintain. 

If this is not the direction you want to take feel free to close this PR. I just did it because it seems like reinventing the wheel a bit.

### DIFF
--- a/tests/test_queue.py
+++ b/tests/test_queue.py
@@ -1,43 +1,42 @@
 import pytest
-from bayes_opt.bayesian_optimization import Queue
+from queue import Queue, Empty
 
 
 def test_add():
     queue = Queue()
 
-    assert len(queue) == 0
-    assert queue.empty
+    assert queue.empty()
 
-    queue.add(1)
-    assert len(queue) == 1
+    queue.put(1)
+    assert queue.qsize() == 1
 
-    queue.add(1)
-    assert len(queue) == 2
+    queue.put(1)
+    assert queue.qsize() == 2
 
-    queue.add(2)
-    assert len(queue) == 3
+    queue.put(2)
+    assert queue.qsize() == 3
 
 
 def test_queue():
 
     queue = Queue()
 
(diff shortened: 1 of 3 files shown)


fmfn:
Of course there is 🤦 .
When implementing it I didn't even think about it, but I agree, not re-inventing the wheel is definitely the way to go. If you don't mind fixing the conflict and squashing the commits into 1, max 2, commits, I would love to get this merged =)

AlbertAlonso:
Ok, first time squashing and it seems I have squashed more that what I should. I will have to check how to fix it, my bad.

AlbertAlonso:
Actually, the Files Changed section of the PR seems about right, but if I check the `use queue from the standard library` commit, it seems to be including some documentation and some other commits. I haven't been able to "fix" it, so maybe it is already ok (?). 
Please let me know what you think. :-)

alonfnt:
Fixed!

tests/test_queue.py
```diff
@@ -1,43 +1,43 @@
...
-from bayes_opt.bayesian_optimization import Queue
+from queue import Queue, Empty
 
 
 def test_add():
     queue = Queue()
 
-    assert len(queue) == 0
+    assert queue.qsize() == 0
     assert queue.empty
```
sonic-chase, line 9: I think this line should be updated to `queue.empty()`

alonfnt: agree :)

bwheelz36:
Hey @alonfnt, sorry about the slow response. Could I ask you to make any commit to this branch, which will trigger the (recently working again CI) to run?

alonfnt:
@bwheelz36 done :)

bwheelz36:
thank you!

PR #240, opened 2020-07-07
Check bounds_transformation is correct type

AlbertAlonso:
A quick check following the 'don't ask for permission ask for forgiveness' motto to make sure that the bounds_transformer used can be called otherwise a clear exception message is shown.

### DIFF
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -87,7 +87,11 @@ class BayesianOptimization(Observable):
         self._verbose = verbose
         self._bounds_transformer = bounds_transformer
         if self._bounds_transformer:
-            self._bounds_transformer.initialize(self._space)
+            try:
+                self._bounds_transformer.initialize(self._space)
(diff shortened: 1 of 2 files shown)


bayes_opt/bayesian_optimization.py
```diff
@@ -87,7 +87,11 @@ def __init__(self, f, pbounds, random_state=None, verbose=2,
         self._verbose = verbose
         self._bounds_transformer = bounds_transformer
         if self._bounds_transformer:
-            self._bounds_transformer.initialize(self._space)
+            try:
+                self._bounds_transformer.initialize(self._space)
+            except (AttributeError, TypeError):
+                raise AssertionError('The transformer must be instance of'
```
fmfn, line 93: Is `AssertionError` the right choice? Maybe `TypeError` instead?

AlbertAlonso: You are right! `TypeError` is a better fit. Fixed.

Issue #300, opened 2022-02-07
Error with scipy 1.8.0


bwheelz36:
I am getting an error with scipy 1.8.0 

```
File "/home/brendan/python/TestVenv/lib/python3.8/site-packages/bayes_opt/util.py", line 65, in acq_max
if max_acq is None or -res.fun[0] >= max_acq:
TypeError: 'float' object is not subscriptable
```

The problem appears to be that in this new version of scipy, res.fun is returned as a number instead of an array. 
I haven't yet found exactly where in scipy the change is. A fix may be as simple as adding an if statement to handle this case. 

I'll keep looking into it but for now I suggest limiting scipy installations to 1.7 or less.

jobquiroz:
I have the same problem. I used   pip install scipy==1.7  and it seems to work.

bfeif:
Came here with the same issue. `pip install scipy==1.7` fixed it for me as well.

amilkaraptuno:
yeah, same issue, fixed with `pip install scipy==1.7`

aeslava97:
Same, `pip install scipy==1.7` did it for me

ngupta20:
Glad to see others have reproduced this issue,

#319 should handle this in a version-explicit way.

Thanks!

bwheelz36:
hey all, just an update on this.
I have permissions to manage this repository now and I have merged #303 into the main branch, which should resolve this error. However, I am still not able to release the code through pypi or conda forge. I will keep trying to get these permissions, but for now I would recommend installing directly from the master branch like this:

```bash
 pip install git+https://github.com/fmfn/BayesianOptimization
```

peterdsharpe:
@bwheelz36 Any update on this? For a package with 6.1k stars, getting this fix pushed to PyPI/conda would seem to be quite critical.

bwheelz36:
Agreed, the problem is I don't actually have the rights to do that yet. I'll try to get in touch with Fernando again though, thanks for the reminder.

bwheelz36:
Assuming I get the rights, I can handle the pypi release - does anyone want to volunteer to help with the conda release? I guess I could figure it out but I don't normally use conda...

xxl4tomxu98:
As far as I am concerned, this has NOT fixed, after I downgraded scipy version to 1.70 and installed "pip install git+https://github.com/fmfn/BayesianOptimization" I still got the same error, can anyone explain why?

xxl4tomxu98:
[...]

PR #303, opened 2022-02-11
Fixes the error caused by scipy 1.8 minimize results changing.

samFarrellDay:
Simple change, just uses `np.squeeze` to ensure the result gets handled as a float.


samFarrellDay:
Fixes #300

charlieexxon (approved):

charlieexxon (approved):

Issue #328, opened 2022-06-17
Issues with SequentialDomainReductionTransformer - one of the lower bounds is greater than an upper bound


MasonBowen:
# High level description of issue
I am wondering how to best identify the cause of and ultimately resolve a persistent error I'm having with:

`ValueError: LBFGSB - one of the lower bounds is greater than an upper bound.`

This happens after toying around a lot with various parameters to no avail.

Not sure what information is helpful and will update based on anything you all say would be helpful in diagnosing. Included here for now is:

- Model Background and inputs
- Some outputs / plot from the real model when it breaks down
- Some plots I have tried to use to track the issue when the model functions properly 
- Toy model that results in similar error
    - error message

Thanks for any help/insight you can spare!!!

## Model Background and inputs
I am not sure which values or what information are most useful for helping identify the problem/solution but will list out a few parameters / information.

- Information
    - Error only happens with the `SequentialDomainReductionTransformer` passed to the `bounds_transformer` of the optimizer (standard optimizer does not throw this error)
    - Error happens even at very low values for init points or when probes are passed
    - the bayes optimizer is optimizing for 14 variables
    - to deal with more nuanced constraints, before a 'guess' is evaluated by the model, it is checked by a separate function which determines if the constraints are violated (e.g., variable_a > variable_b > variable_c). If that constraint is violated a 'spiked' score is returned rather than evaluating the whole model (to be clear this _is_ happening within the optimization function, there's just a "if constraints violated" clause at the beginning that decides whether to pass a score of -20,000 or to run the whole model).

- Parameters:
    - First, I have toyed around with these a lot and still gotten errors, but hopefully this gives an idea of a standard failed run. Please let me know what other information would be useful.
    - To `SequentialDomainReductionTransformer`:
        - `gamma_osc = 0.4`
        - `gamma_pan =  1.0`
        - `eta = 0.99`
    - to `BayesianOptimization`:
        - pbounds: (there are 14 but most look like either of the two listed here:
            - `'res#ElectricityRates#ur_monthly_fixed_charge': (0.0, 50.0)`
            - `'res#ElectricityRates#period_1#rate': (0.05, 0.35)`
        - random_state = `np.random.RandomState(100)`
[...]

PR #332, opened 2022-06-26
Update some of the SequentialDomain transformer 

osullivryan:
The SequentialDomainReduction transformer needed (and still needs) some TLC...

* Only update the domains if you are within the actual searching phase of the optimizer - not when you are pulling from the queue. This causes the bounds to look like they are being broken.
* Add minimum_window for sequential domain reduction so the bounds don't merge to the same value (if the user enters anything)
* If the SequentialDomainReduction tool wants to flip the bounds make sure they are sorted correctly as this can cause an error.


osullivryan:
There is some more refactoring I would like to do to this domain transformer, but I think that should be a different PR.

bwheelz36 (approved):
looks good! and tested that it resolves #328. thanks for adding the new test too!

PR #367, opened 2022-10-11
update version and setup.py

### LINKED ISSUE #362: Bump version
The version of this package has been on 1.2.0 since over two years -- I think with bugs cropping up & getting patched and new features being added, it might be time to increase the version number.

bwheelz36:
- removed setup.cfg, it didnt seem to be doing anything
- remove download_url field for same reason
- updated version such that it is read from the `__init__.__version__` attribute
- bumped version to 1.3.0
- fixes #362

### DIFF
--- a/bayes_opt/__init__.py
+++ b/bayes_opt/__init__.py
@@ -4,8 +4,6 @@ from .util import UtilityFunction
 from .logger import ScreenLogger, JSONLogger
 from .constraint import ConstraintModel
 
-__version__ = '1.3.0'
-
 __all__ = [
     "BayesianOptimization",
     "ConstraintModel",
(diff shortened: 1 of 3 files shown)


bwheelz36:
Ok this turned out to be surpirsingly annoying to update.
@till-m - in my other codes I use a setup.cfg which reads the version from packge.__init__.__version__
I like that because the version is just defined in once place and is clear when importing the package, e.g.
```python
import bayes_opt
print(f'version: {bayes_opt.__version__}')
```
I tried to implement this with setup.py, but was unable to easily do it.
I guess this is why setup.py is now not preferred...

In any case could you have a quick look over this PR

till-m (approved):

Issue #361, opened 2022-10-08
Constraint Error 


qup20:
I would like to ask a question. I used the new version and adding the constraint has no effect, it still fetches the points that are not accepted, why is that? Also, the new version of logger will prompt me: TypeError: Object of type bool_ is not JSON serializable.
Thank you for your answer!

bwheelz36:
Hey, could you please add a minimum example that reproduces these errors so we can debug?

till-m:
The current constrained optimization is designed for constraint functions that are expensive to evaluate (meaning it is not easy to say whether a point fulfills the constraint without explicitly evaluating the constraint function which is computationally costly). In practice, this means that you will sample a lot of points during modelling that do _not_ fulfill the constraints (even by design), because we simply don't know whether a point fulfills the constrain before sampling. On the other hand, we want to make use of the information gained by every evaluation of the constraint function, since this is such a costly operation.
Nonetheless, calling `.max` on the optimizer should give you the point maximizing the function that also fulfills the constraints. 

I hope this clears up the first question to some degree. If you want to read more about the method we are using to model constraints, please read [Gardner et al, 2014](http://proceedings.mlr.press/v32/gardner14.pdf) or feel free to ask a question here.

That being said, if your constraint function is _not_ hard to evaluate, there may be better ways of incorporating your constraints into the optimization. For this, please provide a minimum example.

till-m:
To reproduce the logging bug:
```python
import numpy as np
from bayes_opt import BayesianOptimization
from scipy.optimize import NonlinearConstraint
from bayes_opt.logger import JSONLogger
from bayes_opt.event import Events

logger = JSONLogger(path="./logs.json")

def target_function(x, y):
    # Gardner is looking for the minimum, but this packages looks for maxima, thus the sign switch
    return np.cos(2*x)*np.cos(y) + np.sin(x)

def constraint_function(x, y):
    return np.cos(x) * np.cos(y) - np.sin(x) * np.sin(y)

constraint_limit = 0.5

constraint = NonlinearConstraint(constraint_function, -np.inf, constraint_limit)

# Bounded region of parameter space
pbounds = {'x': (0, 6), 'y': (0, 6)}

optimizer = BayesianOptimization(
    f=target_function,
    constraint=constraint,
    pbounds=pbounds
    random_state=1,
)
optimizer.subscribe(Events.OPTIMIZATION_STEP, logger)
optimizer.maximize(
    init_points=2,
[...]

Issue #353, opened 2022-09-12
replace non-unique point error with warning?


bwheelz36:
As has been noted in #158, the code will crash if it is asked to probe a point it has already probed.

This condition occurs not infrequently when using the utility functions. In addition, when working with very noisy data, it is completely valid to probe the same point more than once.

**Describe the solution you'd like**
I suggest simply replacing the error with a warning, and maybe a parameter called '_n_repeated_points_probed' - that would allow users to handle the issue on their end. Thoughts?

**Are you able and willing to implement this feature yourself and open a pull request?**
- [x] Yes, I can provide this feature.

till-m:
Two (naive) questions that I have:

> In addition, when working with very noisy data, it is completely valid to probe the same point more than once.

How do the GPR's handle receiving different target values for the same parameters? Does it just take the mean? Are the covariances adjusted at that point (or for all the points)?

> a parameter called '_n_repeated_points_probed'

I assume -- based on the name -- that this is an integer counter storing how often points that were already probed have been re-probed?

bwheelz36:
Hey @till-m 

> "I assume -- based on the name -- that this is an integer counter storing how often points that were already probed have been re-probed?"

correct!

> "How do the GPR's handle receiving different target values for the same parameters? Does it just take the mean? Are the covariances adjusted at that point (or for all the points)?"

This is actually something I am currently trying to figure out. I've been playing around with 10 runs of the same simulation, with varying levels of noise and seeing how best to model it. The distribution of the different data sets is below (n here means n particles, but that's not really important for the purpose of this discussion).
![Figure_1](https://user-images.githubusercontent.com/64300337/189774374-d4b4c18a-c2e8-4602-b361-6ada21322b33.png)

If we run the bayesian_optimization code with the default parameters, repeatedly pass it the same point with different (noisy) values, and simply replace the `raise repeated_points` error with a pass statement, it runs without issue but doesn't do a good job handling the noise:

![Figure_1](https://user-images.githubusercontent.com/64300337/189775112-e435ffef-de48-456a-8b11-7555d5561522.png)

However, we can construct a custom kernel like this:

```python
from bayes_opt import BayesianOptimization
from sklearn.gaussian_process.kernels import Matern, WhiteKernel

optimizer = BayesianOptimization(f=None,pbounds=pbounds,  verbose=2, random_state=1)

k1 = Matern(length_scale=[3, 0.2, 0.2])  # Matern is the defaul kernel
[...]

PR #368, opened 2022-10-13
Improve constrained optimization

### LINKED ISSUE #363: Logging with Constraints: `ScreenLogger` doesn't print information related to constraints
(Less of a bug and more of an oversight on my part)

**Describe the bug**
When using `ScreenLogger`, no information about the constrained optimization is printed.

**To Reproduce**

```python
import numpy as np
from scipy.optimize import NonlinearConstraint

def target_function(x, y):
    return np.cos(2*x)*np.cos(y) + np.sin(x)

def constraint_function(x, y):
    return np.cos(x) * np.cos(y) - np.sin(x) * np.sin(y)

constraint_limit = 0.5
constraint = NonlinearConstraint(constraint_function, -np.inf, constraint_limit)

# Bounded region of parameter space
pbounds = {'x': (0, 6), 'y': (0, 6)}

    f=target_function,
    constraint=constraint,
    pbounds=pbounds,
    random_state=1,
)

optimizer.maximize(
    init_points=2,
    n_iter=10,
)
```

**Expected behavior**
I'm somewhat torn on what information should be output. I see a few options:
- Just the value(s) of the constraint function(s) at the specific point
- Just the `"allowed"`-ness of the point
- Both the specific values and the `"allowed"`-ness
- Colour disallowed points red and/or allowed points green
- Just the value(s) of the constraint function(s) at the specific point, colour coded green/red for allowed/disallowed 

Personally I'm leaning towards the second option to reduce clutter, but I'm open to other opinions. It would look something like this:

```
|   iter    |  target   |  allowed  |     x     |     y     |
-------------------------------------------------------------
| 1         | 0.4872    | False     | 2.502     | 4.322     |
| 2         | -0.2401   | True      | 0.0006862 | 1.814     |
| 3         | 0.7821    | False     | 2.304     | 4.326     |
| 4         | 1.249     | False     | 1.899     | 4.321     |
| 5         | 0.7351    | False     | 1.351     | 4.982     |
| 6         | 1.814     | True      | 1.534     | 3.757     |
| 7         | 1.793     | False     | 1.854     | 2.982     |
| 8         | 0.02935   | False     | 2.834     | 1.913     |
| 9         | 1.496     | False     | 1.108     | 3.106     |
| 10        | 0.5423    | True      | 5.99      | 0.06449   |
| 11        | 0.4597    | True      | 5.985     | 5.858     |
| 12        | -1.095    | True      | 5.986     | 2.891     |
=============================================================
```

till-m:
- Fix #363: ScreenLogger now prints "allowed"-ness
- Assert that `lb` (lower bound) is lower than `ub` upper bound
    Something that I mentioned in the original PR but then forgot to do.
- Merge `ConstrainedTargetSpace` into `TargetSpace`
Originally I thought there would be less clutter if I subclass the original `TargetSpace`, but I think it will make future maintenance much harder since code will need to be changed in multiple places at once (see e.g. #365).
- Fix a bug related to one-dimensional constraints with bounds of type `int` (presumably the bug mentioned in #361)
- Allow loading of logs with constraint information (see #363 as well)
- Mention the constrained optimization notebook in the quickstart section of the README.
- Added a stopgap to deal with registering a duplicate point when loading logs.


bwheelz36:
I have a feeling that test may have failed randomly, can you make a tiny change and push to force it to run again?
This failure would be fixed when I address #353

till-m:
I added a stopgap so that I can make sure loading from logs behaves correctly and replaced the (imho) slightly obscure `KeyError` that was raised. 

@bwheelz36 feel free to review at your convenience -- I've addressed all the things I wanted to address.

Issue #208, opened 2020-02-21
Formatting issues in Windows


LucaCerina:
Hello,
when executed in Windows, the optimization have some weird formatting errors. Like this:

|   iter    |  target   |   bits    |   size    |
-------------------------------------------------
| [0m 1       [0m | [0m 1.0     [0m | [0m 21.34   [0m | [0m 1.476e+0[0m |
| [0m 2       [0m | [0m 1.0     [0m | [0m 8.004   [0m | [0m 620.6   [0m |
| [0m 3       [0m | [0m 1.0     [0m | [0m 12.7    [0m | [0m 190.9   [0m |
| [0m 4       [0m | [0m 1.0     [0m | [0m 13.96   [0m | [0m 709.0   [0m |
| [0m 5       [0m | [0m 1.0     [0m | [0m 20.7    [0m | [0m 1.104e+0[0m |
| [0m 6       [0m | [0m 1.0     [0m | [0m 21.41   [0m | [0m 1.404e+0[0m |
| [0m 7       [0m | [0m 1.0     [0m | [0m 14.54   [0m | [0m 1.799e+0[0m |
| [0m 8       [0m | [0m 1.0     [0m | [0m 8.876   [0m | [0m 1.374e+0[0m |
| [0m 9       [0m | [0m 1.0     [0m | [0m 21.35   [0m | [0m 1.145e+0[0m |
| [0m 10      [0m | [0m 1.0     [0m | [0m 12.49   [0m | [0m 407.3   [0m |

MohamedNabil97:
I have the same issue

fmfn:
Oh, I believe the way [colours are implemented](https://github.com/fmfn/BayesianOptimization/blob/af48b0374fc73851d34818d410b01e08eedb1e07/bayes_opt/util.py#L178) is not compatible with Windows.

Unfortunately I don't have the bandwidth to fix this currently (nor do I have a way of testing it), so PRs are more than welcome.

MohamedNabil97:
@fmfn  I heard about a python module called colorama that may help fixing this issue. https://pypi.org/project/colorama/ I haven't personally tried it out yet but it seems promising.

fmfn:
Yep, colorama will definitely do the trick, and it shouldn't be too much work either. Still, I don't plan on tackling this anytime soon.

m-zanichelli:
Hi, did someone succeed in solving this issue? I'm currently having similar problems

bwheelz36:
[...]

PR #372, opened 2022-10-27
Fix duplicate points

bwheelz36:


bayes_opt/target_space.py
```diff
@@ -176,8 +181,12 @@ def register(self, params, target, constraint_value=None):
         """
         x = self._as_array(params)
         if x in self:
-            raise NotUniqueError('Data point {} is not unique'.format(x))
-
+            if self._allow_duplicate_points:
+                self.n_duplicate_points = self.n_duplicate_points + 1
+                warnings.warn(f'Data point {x} is not unique. {self.n_duplicate_points} duplicates registered.'
```
till-m, line 186: FYI warnings are usually only shown once, so this will not give you status updates about the number of duplicates.

bayes_opt/target_space.py
```diff
@@ -37,6 +41,8 @@ def __init__(self, target_func, pbounds, constraint=None, random_state=None):
             optionally specify a seed for a random number generator
         """
         self.random_state = ensure_rng(random_state)
+        self._allow_duplicate_points = allow_duplicate_points
```
till-m, line 44: I think this should be explicitly mutable, so I would not prepend the underscore.

bayes_opt/target_space.py
bwheelz36: I can't believe I had never noticed this!

bayes_opt/target_space.py
bwheelz36: you think we should enable this to be changed post instantiation? 
since the behavior is inside `Optimizer._space` which is itself private, then would have to add some sort of setting method to the main `Optimizer` method?

till-m: Fair enough. Let's leave it like this then.

bayes_opt/target_space.py
till-m: This will probably break the output of the `ScreenLogger`, right?
I think this way is perfectly fine, but if you want to use warnings & have them print repeatedly that would be possible by adding context, see [here](https://stackoverflow.com/a/22661838/12051450).

till-m: As a sidenote regarding colours -- do you know if #208 has been fixed or not?

bwheelz36: yeah I guess it won't look great with screen logger.
perhaps a more elegant and appropriate way to do this is actually make the probing of duplicate points an event that can then get logged?

bwheelz36: At least for me I have no problems with the screen logger on windows so I closed #208.

till-m: > perhaps a more elegant and appropriate way to do this is actually make the probing of duplicate points an event that can then get logged?

That makes sense. Still, I'm having trouble picturing how this would like in the context of the `ScreenLogger` output. Do you have a specific idea?

bwheelz36: Hey @till-m,
I thought about this a bit more.
Although the current approach does 'ruin' the output of the screen logger, I think that in most situations the screen logger should not be relied on anyway. I contemplated adding an extra column 'duplicate point (True/False)' but this seemed like overkill to address a situation that hopefully is not actually occurring that often anyway. 
Similarly, regarding making this an 'event' - to be honest this aspect of the code confuses me somewhat. From my perspective, seeing that the duplicate points are already clearly defined in the logs I'm not sure any more is needed...
Happy to hear other perspectives/approaches!

bayes_opt/bayesian_optimization.py
```diff
@@ -62,6 +62,8 @@ def dispatch(self, event):
             callback(event, self)
 
 
+
```
till-m, line 65: Is there a reason for adding these newlines?

bayes_opt/bayesian_optimization.py
```diff
@@ -242,12 +246,8 @@ def _prime_subscriptions(self):
...
                  init_points=5,
                  n_iter=25,
-                 acq='ucb',
-                 kappa=2.576,
-                 kappa_decay=1,
-                 kappa_decay_delay=0,
-                 xi=0.0,
-                 **gp_params):
+                 acquisition_function=None):
+
```
till-m, line 250: docstring outdated.

examples/test.py
```diff
@@ -0,0 +1,53 @@
...
+        utility = UtilityFunction(kind="ucb", kappa=5, xi=1)  # kappa determines explore/Exploitation ratio
+        # optimizer.maximize(
+        #     init_points=2,
+        #     n_iter=20,
+        # )
+
+        optimizer.maximize(init_points=2,
+                     n_iter=25,
+                     utility_function=utility)
+        # for point in range(20):
```
till-m, line 31: any particular reason to keep this commented code?

bayes_opt/target_space.py
```diff
@@ -22,7 +25,9 @@ class TargetSpace(object):
     >>> y = space.register_point(x)
     >>> assert self.max_point()['max_val'] == y
     """
-    def __init__(self, target_func, pbounds, constraint=None, random_state=None):
+
+    def __init__(self, target_func, pbounds, constraint=None, random_state=None,
```
till-m, line 29: docstring outdated.

bayes_opt/bayesian_optimization.py
```diff
@@ -111,9 +113,10 @@ def __init__(self,
                  constraint=None,
                  random_state=None,
                  verbose=2,
-                 bounds_transformer=None):
+                 bounds_transformer=None,
```
till-m, line 116: docstring outdated.

till-m (approved):
looks good to me

bayes_opt/target_space.py
till-m: I agree. Probably not worth worrying about. I would ship this as-is.

PR #383, opened 2022-11-25
Update simple contraints docs

bwheelz36:
just adding docs on the case of simple constraints


till-m:
FYI https://twitter.com/hynek/status/1596039449909264385

examples/constraints.ipynb
```diff
@@ -1,5 +1,149 @@
 {
  "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Constrained Optimization\n",
+    "\n",
+    "Constrained optimization refers to situations in which you must for instance maximize `f`, a function of `x` and `y`, but the solution must lie in a region where for instance `x<y`.\n",
```
till-m, line 9: these don't render properly for me (see image). I would suggest using TeX here.
Screenshot 2022-11-25 at 15 44 05

bwheelz36: how odd. I will just remove the backticks I guess.
I just noticed there's quite a lot of other typos that need fixing too!

PR #388, opened 2022-12-02
Add python version requirement, colorama version requirement

till-m:
fix #386 
kinda-fixes #387

PR #389, opened 2022-12-02
Note minimum colorama version

akx:
`colorama.just_fix_windows_console` was added in Colorama 0.4.6, so that's the minimum version.

(See https://stackoverflow.com/q/74654425/51685 for context.)


bwheelz36:
thanks, fixed in #388

Issue #405, opened 2023-04-05
BayesianOptimization.maximize behaves unexpectedly when passing any of the deprecated parameters, in a non-backward-compatible way


godot11:
In a (probably recent) major API change, passing the `acq, kappa, kappa_decay, kappa_decay_delay, xi` parameters to `BayesianOptimization.maximize` apparently became deprecated.

However, right now, `BayesianOptimization.maximize` accepts these parameters, and **silently TOTALLY ignores them**. Only a DeprecationWarning is emitted, (which is silent by default unless the function is in the top library), as seen in the relevant lines in `bayesian_optimization`:

```python
    old_params_used = any([param is not None for param in [acq, kappa, kappa_decay, kappa_decay_delay, xi]])
        if old_params_used or gp_params:
            warnings.warn('\nPassing acquisition function parameters or gaussian process parameters to maximize'
                                     '\nis no longer supported, and will cause an error in future releases. Instead,'
                                     '\nplease use the "set_gp_params" method to set the gp params, and pass an instance'
                                     '\n of bayes_opt.util.UtilityFunction using the acquisition_function argument\n',
                          DeprecationWarning, stacklevel=2)

        if acquisition_function is None:
            util = UtilityFunction(kind='ucb',
                                   kappa=2.576,
                                   xi=0.0,
                                   kappa_decay=1,
                                   kappa_decay_delay=0)
        else:
            util = acquisition_function
```

This behavior does not make sense. It is absolutely NOT backwards compatible, and worse, breaks existing code in a very confusing, unexpected, and hard-to-debug way. Backwards compatibility doesn't mean that the change won't cause syntax error in existing code, it means that existing code will function as before. 

In this case, either the values should be passed to the `UtilityFunction`, or a `ValueError` should be emitted.

In fact, this way of implementation caused our team to lose a week worth of effort, as we wanted to compare the three implemented utility function, and the effect of different kappa and kappa-decay values on `'ucb'`. We were puzzled why we don't see any variation like previously using the same code. As we call `BayesianOptimization.maximize` from within a homebrew library, the warning was never printed, everything seemed to be fine, and the code just silently replaced all of our parameters with the default values.

Even if the warning *was* emitted, however, it doesn't tell that the values are *already being ignored*. It's just a warning that this API is deprecated and will be turned into an error. 

[...]

PR #406, opened 2023-04-12
change Deprecation warning to Error

bwheelz36:
address issues raised in #405


bwheelz36:
ok turns out the coverage approach is deprecated, will fix in a different PR

till-m (approved):

PR #413, opened 2023-05-05
Update acquisition function scaling for constrained optimization

till-m:
This makes the acqusition function equivalent to the unconstrained case if the GP is certain that the constraint is fulfilled.

NB: Should we add a note somewhere that constrained optimization is best used with the EI acquisition function?

### DIFF
--- a/bayes_opt/util.py
+++ b/bayes_opt/util.py
@@ -69,7 +69,7 @@ def acq_max(ac, gp, y_max, bounds, random_state, constraint=None, n_warmup=10000
             if target < 0:
                 return target * p_constraint
             else:
-                return target / (0.5 + p_constraint)
+                return target / (0.5 + 0.5 * p_constraint)
     else:
         to_minimize = lambda x: -ac(x.reshape(1, -1), gp=gp, y_max=y_max)
 


bwheelz36:
Hi @till-m , I don't 100% understand this but all checks are passing and it seems like you do, so from this perspective I'm happy to approve :-) you can merge this yourself now right?

bwheelz36 (approved):

till-m:
> Hi @till-m , I don't 100% understand this but all checks are passing and it seems like you do, so from this perspective I'm happy to approve :-) you can merge this yourself now right?

Fair point, I'll just outline my thinking for the record:
Gardner derived constrained optimization while using EI, which is non-negative. However, if an acquisition function is used that is not non-negative we cannot just scale by probability (this would make negative acq-value points more attractive if the model is sure of the constraint not being fulfilled!). Because of that, I thought up some sort of scaling law for negative values of the acquisition function when I made the original PR. The idea here was to have the same properties that of the positive case:
1) Make exploring values more attractive if the model is rather sure of the constraint being fulfilled.
2) Make exploring values less attractive if the model is rather sure of the constraint not being fulfilled.

However, the original formulation by Gardner has a third nice property, which is leaving the objective unchanged (compared to the unconstrained optimization) if the model is sure of the constraint being fulfilled. This probably makes little difference in practice, but I felt it would maybe be nice to have this property in the negative case as well, thus the small change in rescaling.

PR #415, opened 2023-05-05
Partially fix #412 by adding a dedicated method to evaluating y_max

### LINKED ISSUE #412: Constrained optimization
Hello,

I was looking into the constrained optimization code path and noticed a few things and I would like to ask some questions to see if my understanding is correct or not.

This is a snippet from the "acq_max" function:

```python
def acq_max(ac, gp, y_max, bounds, random_state, constraint=None, n_warmup=10000, n_iter=10):
    x_tries = random_state.uniform(bounds[:, 0], bounds[:, 1],
                                   size=(n_warmup, bounds.shape[0]))
    ys = ac(x_tries, gp=gp, y_max=y_max)
    x_max = x_tries[ys.argmax()]
    max_acq = ys.max()

    # Explore the parameter space more throughly
    x_seeds = random_state.uniform(bounds[:, 0], bounds[:, 1],
                                   size=(n_iter, bounds.shape[0]))

    if constraint is not None:
        def to_minimize(x):
            target = -ac(x.reshape(1, -1), gp=gp, y_max=y_max)
            p_constraint = constraint.predict(x.reshape(1, -1))

            # TODO: This is not exactly how Gardner et al do it.
            # Their way would require the result of the acquisition function
            # to be strictly positive (or negative), which is not the case
            # here. For a negative target value, we use Gardner's version. If
            # the target is positive, we instead slightly rescale the target
            # depending on the probability estimate to fulfill the constraint.
            if target < 0:
                return target * p_constraint
            else:
                return target / (0.5 + p_constraint)
    else:
        to_minimize = lambda x: -ac(x.reshape(1, -1), gp=gp, y_max=y_max)

    for x_try in x_seeds:
        # Find the minimum of minus the acquisition function
        res = minimize(lambda x: to_minimize(x),
                       x_try,
                       bounds=bounds,
                       method="L-BFGS-B")

        # See if success
        if not res.success:
            continue

        # Store it if better than previous minimum(maximum).
        if max_acq is None or -np.squeeze(res.fun) >= max_acq:
            x_max = res.x
            max_acq = -np.squeeze(res.fun)

    return np.clip(x_max, bounds[:, 0], bounds[:, 1])
```

The first thing that I noticed was that the `y_max` that is passed here is the "unconstrained y_max" and not the "constrained y_max" and in the original paper I can cite this:

> Thus the expected constrained improvement acquisition
function EIC (xˆ) is precisely the standard expected improvement of xˆ over the best feasible point so far weighted
by the probability that xˆ is feasible. 

The second thing is that there is a "warm up" phase with cheap evaluations, this phase is not considering the `p_constraint` at all, which mean that they will get higher `y_max` values than normal (the same as p_constraint = 1).
```python
    x_tries = random_state.uniform(bounds[:, 0], bounds[:, 1],
                                   size=(n_warmup, bounds.shape[0]))
    ys = ac(x_tries, gp=gp, y_max=y_max)
    x_max = x_tries[ys.argmax()]
    max_acq = ys.max()
```

Is this a bug? Does this make sense? I'm no expert in the area.

If it makes sense, I can help with a PR here.

Thank you!

leandrobbraga:
Partially Fixes #412.

The idea is to have a dedicated space property that calculates eagerly the `y_max` of the space, considering the constraints.

I decided to split the MR in two to ease the review. I'll be sending another MR soon with the fix of the warmup phase of the `acq_max`.

### DIFF
--- a/bayes_opt/target_space.py
+++ b/bayes_opt/target_space.py
@@ -265,44 +265,54 @@ class TargetSpace(object):
             data.T[col] = self.random_state.uniform(lower, upper, size=1)
         return data.ravel()
 
+    def y_max(self):
+        """Get maximum target value found.
+        
+        If there is a constraint present, the maximum value that fulfills the
+        constraint is returned."""
+        if len(self.target) == 0:
+            return None
+
+        if self._constraint is None:
+            return self.target.max()
+
+        allowed = self._constraint.allowed(self._constraint_values)
+        if allowed.any():
+            return self.target[allowed].max()
+
+        return None
+
     def max(self):
         """Get maximum target value found and corresponding parameters.
         
         If there is a constraint present, the maximum value that fulfills the
         constraint is returned."""
-        if self._constraint is None:
-            try:
-                res = {
-                    'target': self.target.max(),
-                    'params': dict(
-                        zip(self.keys, self.params[self.target.argmax()])
-                    )
-                }
-            except ValueError:
-                res = {}
+        y_max = self.y_max()
+
+        if y_max is None:
+            res = {
+                'target': None,
+                'params': None
+            }
+
+            if self._constraint is not None:
+                res['constraint'] = None
+
             return res
-        else:
-            allowed = self._constraint.allowed(self._constraint_values)
-            if allowed.any():
-                # Getting of all points that fulfill the constraints, find the
-                # one with the maximum value for the target function.
-                sorted = np.argsort(self.target)
-                idx = sorted[allowed[sorted]][-1]
-                # there must be a better way to do this, right?
-                res = {
-                    'target': self.target[idx],
-                    'params': dict(
-                        zip(self.keys, self.params[idx])
-                    ),
-                    'constraint': self._constraint_values[idx]
-                }
-            else:
-                res = {
-                    'target': None,
-                    'params': None,
-                    'constraint': None
-                }
-            return res
+
+        y_idx = np.where(self.target == y_max)[0][0]
+
+        res = {
+                'target': y_max,
+                'params': dict(
+                zip(self.keys, self.params[y_idx])
+            )
+        }
+
+        if self._constraint is not None:
+            res['constraint'] = self._constraint_values[y_idx]
+
+        return res
 
     def res(self):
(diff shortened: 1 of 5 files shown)


bayes_opt/target_space.py
```diff
@@ -265,45 +265,55 @@ def random_sample(self):
...
+        y_max = self.y_max()
+
+        if y_max is None:
+            res = {
+                'target': None,
+                'params': None
+            }
+
+            if self._constraint is not None:
+                res['constraint'] = None
```
leandrobbraga, line 299: I guess this is a breaking change, if you wan't me to keep the old behaviour I can add another commit.

I did it like this because in the old behaviour we had a `res = {}` for unconstrained case and `res = {'target': None, ]params': None, 'constraint': None}` for the constrained case and I wanted to unify the behaviour.

till-m (commented):
Looks good to me in general. Small note, as far as I know we're generally using `target` instead of `y` for the result of the function we are maximizing. I know that this is not consistent everywhere, but I think it would be good to go that direction here.

bayes_opt/target_space.py
```diff
@@ -265,45 +265,55 @@ def random_sample(self):
             data.T[col] = self.random_state.uniform(lower, upper, size=1)
         return data.ravel()
 
+    def y_max(self):
```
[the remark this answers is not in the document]
bwheelz36: anything not user facing should have `_`. I see it is `_target_space` now so all good :-)

bayes_opt/target_space.py
bwheelz36: I agree - both that None is better and also that either is ok :-). This still seems to be returning a dict of None at the moment?

bwheelz36 (approved):

bayes_opt/target_space.py
till-m: Let's go with `None` then. That should be the last thing and then we can merge this!

leandrobbraga: Done

leandrobbraga:
I had to adapt the tests after the last change, some were failing

till-m:
Thanks for your contribution!

PR #416, opened 2023-05-05
Fix warmup phase not considering constraints

### LINKED ISSUE #412: Constrained optimization

leandrobbraga:
Fixes #412 

Previously the warmup phase would use the acquisition function directly without considering the p_constraint. This could lead to inflated values of the acquisition function in unfeasible spaces.

### DIFF
--- a/bayes_opt/util.py
+++ b/bayes_opt/util.py
@@ -46,14 +46,16 @@ def acq_max(ac, gp, y_max, bounds, random_state, constraint=None, n_warmup=10000
 
     # We need to adjust the acquisition function to deal with constraints when there is some
     if constraint is not None:
-
-        def adjusted_ac(x, gp, y_max):
+        def adjusted_ac(x):
             """Acquisition function adjusted to fulfill the constraint when necessary"""
-            values = ac(x, gp=gp, y_max=y_max)
-            p_constraints = constraint.predict(x)
+
+            # Transforms the problem in a minimization problem, this is necessary
+            # because the solver we are using later on is a minimizer
+            values = -ac(x.reshape(-1, bounds.shape[0]), gp=gp, y_max=y_max)
+            p_constraints = constraint.predict(x.reshape(-1, bounds.shape[0]))
 
             # Slower fallback for the case where any values are negative
-            if np.any(values < 0):
+            if np.any(values > 0):
                 # TODO: This is not exactly how Gardner et al do it.
                 # Their way would require the result of the acquisition function
                 # to be strictly positive, which is not the case here. For a
@@ -62,7 +64,7 @@ def acq_max(ac, gp, y_max, bounds, random_state, constraint=None, n_warmup=10000
                 # on the probability estimate to fulfill the constraint.
                 return np.array(
                     [
-                        value * p if value > 0 else value / (0.5 + 0.5 * p)
+                        value / (0.5 + 0.5 * p) if value > 0 else value * p
                         for value, p in zip(values, p_constraints)
                     ]
                 )
@@ -71,12 +73,14 @@ def acq_max(ac, gp, y_max, bounds, random_state, constraint=None, n_warmup=10000
             return values * p_constraints
 
     else:
-        adjusted_ac = ac
+        # Transforms the problem in a minimization problem, this is necessary
+        # because the solver we are using later on is a minimizer
+        adjusted_ac = lambda x: -ac(x.reshape(-1, bounds.shape[0]), gp=gp, y_max=y_max)
 
     # Warm up with random points
     x_tries = random_state.uniform(bounds[:, 0], bounds[:, 1],
                                    size=(n_warmup, bounds.shape[0]))
-    ys = adjusted_ac(x_tries, gp=gp, y_max=y_max)
+    ys = -adjusted_ac(x_tries)
     x_max = x_tries[ys.argmax()]
     max_acq = ys.max()
 
@@ -86,7 +90,7 @@ def acq_max(ac, gp, y_max, bounds, random_state, constraint=None, n_warmup=10000
 
     for x_try in x_seeds:
         # Find the minimum of minus the acquisition function
-        res = minimize(lambda x: -adjusted_ac(x.reshape(1, -1), gp=gp, y_max=y_max),
+        res = minimize(adjusted_ac,
                        x_try,
                        bounds=bounds,
                        method="L-BFGS-B")
--- a/bayes_opt/util.py
+++ b/bayes_opt/util.py
@@ -44,10 +44,39 @@ def acq_max(ac, gp, y_max, bounds, random_state, constraint=None, n_warmup=10000
     :return: x_max, The arg max of the acquisition function.
     """
 
+    # We need to adjust the acquisition function to deal with constraints when there is some
+    if constraint is not None:
+
+        def adjusted_ac(x, gp, y_max):
+            """Acquisition function adjusted to fulfill the constraint when necessary"""
+            values = ac(x, gp=gp, y_max=y_max)
+            p_constraints = constraint.predict(x)
+
+            # Slower fallback for the case where any values are negative
+            if np.any(values < 0):
+                # TODO: This is not exactly how Gardner et al do it.
+                # Their way would require the result of the acquisition function
+                # to be strictly positive, which is not the case here. For a
+                # positive target value, we use Gardner's version. If the target
+                # is negative, we instead slightly rescale the target depending
+                # on the probability estimate to fulfill the constraint.
+                return np.array(
+                    [
+                        value * p if value > 0 else value / (0.5 + p)
+                        for value, p in zip(values, p_constraints)
+                    ]
+                )
+
+            # Faster, vectorized version of Gardner et al's method
+            return values * p_constraints
+
+    else:
+        adjusted_ac = ac
+
     # Warm up with random points
     x_tries = random_state.uniform(bounds[:, 0], bounds[:, 1],
                                    size=(n_warmup, bounds.shape[0]))
-    ys = ac(x_tries, gp=gp, y_max=y_max)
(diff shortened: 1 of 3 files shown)


tests/test_constraint.py
```diff
@@ -147,7 +147,7 @@ def constraint_function_2_dim(x, y):
     params = optimizer.res[0]["params"]
     x, y = params['x'], params['y']
 
-    assert constraint_function_2_dim(x, y) == approx(optimizer.constraint.approx(np.array([x, y])), rel=1e-5, abs=1e-5)
+    assert constraint_function_2_dim(x, y) == approx(optimizer.constraint.approx(np.array([x, y])), rel=1e-3, abs=1e-3)
```
leandrobbraga, line 150: I did this adjustment here, do you think it is fine?

Is there any new test I could add that you can think of?

bayes_opt/util.py
```diff
@@ -44,38 +44,49 @@ def acq_max(ac, gp, y_max, bounds, random_state, constraint=None, n_warmup=10000
...
+                # to be strictly positive, which is not the case here. For a
+                # positive target value, we use Gardner's version. If the target
+                # is negative, we instead slightly rescale the target depending
+                # on the probability estimate to fulfill the constraint.
+                return np.array(
+                    [
+                        value * p if value > 0 else value / (0.5 + 0.5 * p)
+                        for value, p in zip(values, p_constraints)
+                    ]
+                )
```
leandrobbraga, line 68: We could vectorize this as well (much much faster) if we can ensure that always all values have the same sign (positive or negative). Is the sign a property of the acquisition function?

tests/test_constraint.py
till-m: > Is there any new test I could add that you can think of?

I think you're good! Coverage is currently increasing with this PR. If you want a closer look, you can see [the report](https://app.codecov.io/gh/fmfn/BayesianOptimization/pull/415/tree/bayes_opt), some lines are only covered partially, but none of is an issue, imo.

till-m (approved):

till-m (approved):

bayes_opt/util.py
```diff
@@ -44,38 +44,49 @@ def acq_max(ac, gp, y_max, bounds, random_state, constraint=None, n_warmup=10000
     :return: x_max, The arg max of the acquisition function.
     """
 
+    # We need to adjust the acquisition function to deal with constraints when there is some
+    if constraint is not None:
+
+        def adjusted_ac(x, gp, y_max):
```
bwheelz36, line 50: 1. this is a somewhat minor point, but this pattern seems confusing - there is no reason for `adjusted_ac` to be inside `acq_max` is there? perhaps this is just a preference... 
2. I think we only ever intend these functions to be called from inside the code (right @till-m?) - in that case should we take this chance to append the `_` to them?

bwheelz36 (approved):

bayes_opt/util.py
leandrobbraga: Done

till-m: 2\. is a consequence of 1.: The function is only defined within the scope of acq max so it cannot be accessed from outside, in that sense prepending the `_` is not necessary.

Since we need to provide `minimize` with a single-argument function, we need to define a function locally (either named, like this, or anonymous, using `lambda`). This function is also only relevant for this specific optimization step -- earlier or later the information that we're working with (represented by `gp` and `y_max`) will have changed, and so will the function to optimize. In this sense, I think this function should indeed be defined here and also doesn't need an underscore. However, the two-layer usage that is introduced in this PR is a bit suboptimal IMO: First, you define `adjusted_ac` and then later you wrap that in a lambda function. Since `y_max` and `gp` don't change during the `acq_max` step, they shouldn't be used as inputs for the function and the construction of the lambda function will not be necessary.

leandrobbraga: I see what you mean, but there is a few issues of removing the `lambda` in the `minimize`:

1) The warmup phase is a maximization problem and the second phase is a minimization problem, we could unify this behaviour and transform both problems in a minimization problem. ✅ 

2) To remove the `gp` and `y_max` from the function we would need to introduce a new lambda/function in the "else" part

```python
    if constraint is not None:

        def adjusted_ac(x, gp, y_max):
            (...)
            return values * p_constraints
    else:
        adjusted_ac = lambda x: -ac(x, gp=gp, y_max=y_max)
```

I guess one advantage is that the `minimize` part do many calls and the warmup phase does only a single call.

3) We would also need to conditionally `.reshape(1, -1` inside the `adjusted_ac`, because the `minimize` passes a 1D array if I'm not mistaken.

```python
    if constraint is not None:

        def adjusted_ac(x, gp, y_max):
            """Acquisition function adjusted to fulfill the constraint when necessary"""
            if len(x.shape) == 1:
                x = x.reshape(1, -1)

            values = ac(x, gp=gp, y_max=y_max)
            (...)
            return values * p_constraints

    else:
        adjusted_ac = lambda x: -ac(x.reshape(1, -1) if len(x.shape) == 1 else x, gp=gp, y_max=y_max)
```

What do you guys prefer?

leandrobbraga: I removed the commit that added the `_` to the `adjusted_ac`  function.

till-m: Something like this should work, right?
```python
    if constraint is not None:
        def adjusted_ac(x):
            """Acquisition function adjusted to fulfill the constraint when necessary"""
            values = ac(x.reshape((-1, bounds.shape[0]), gp=gp, y_max=y_max)
            p_constraints = constraint.predict(x)

            # Slower fallback for the case where any values are negative
            if np.any(values < 0):
                # TODO: This is not exactly how Gardner et al do it.
                # Their way would require the result of the acquisition function
                # to be strictly positive, which is not the case here. For a
                # positive target value, we use Gardner's version. If the target
                # is negative, we instead slightly rescale the target depending
                # on the probability estimate to fulfill the constraint.
                return np.array(
                    [
                        value * p if value > 0 else value / (0.5 + 0.5 * p)
                        for value, p in zip(values, p_constraints)
                    ]
                )

            # Faster, vectorized version of Gardner et al's method
            return values * p_constraints

    else:
        adjusted_ac = lambda x: ac(x.reshape((-1, bounds.shape[0]), gp=gp, y_max=y_max)
```

Could also think of assigning `bounds.shape[0]` to a meaningfully named variable since it's reused a few times in this section.

leandrobbraga: Did what your suggested.

I couldn't find a good name for `bounds.shape[0]` do you have a suggestion?

till-m:
PR looks good to me! Will also merge tomorrow, provided merging #415 doesn't cause any conflicts!

tests/test_constraint.py
till-m: @leandrobbraga can you confirm that this test fails now?

leandrobbraga: The difference starts in the 4th decimal place

image

till-m: thanks!

till-m:
Thanks @leandrobbraga, both for spotting the problems in the first place and for the excellent contributions you've made!

PR #417, opened 2023-05-07
Add run notebooks test

bwheelz36:
I've added a test that simply runs through all the notebooks in the examples folder. the test if fails if any notebook raises an error.


examples/constraints.ipynb
```diff
@@ -183,9 +183,21 @@
...
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 4,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "ename": "ModuleNotFoundError",
+     "evalue": "No module named 'matplotlib'",
+     "output_type": "error",
```
till-m, line 192: There seems to be an error here.

bwheelz36:
Hi @till-m, we know it works because it has just run over this pull request - if you go into one of the test logs you will see that 'test_notebooks_run' has been executed and is passing.
The error you found (good catch!) is not in the tests, but me trying and failing to run the notebook and accidentally saving the result. I will push a fix for that in a minute.
The tests won't actually edit the existing notebooks (this is possible but painful with version control) but merely ensure that they can execute.

till-m (approved):
Okay, this seems good to me.

> The tests won't actually edit the existing notebooks (this is possible but painful with version control) but merely ensure that they can execute.

I think this is fine. PR looks good to me! 👍 

If a notebook is failing, is it obvious in which notebook the failure occured from the test logs?

till-m:
Minor nitpick, but is it necessary to change the notebooks in this PR? I anticipate that there might be some merge conflicts with #419.

bwheelz36:
> " If a notebook is failing, is it obvious in which notebook the failure occured from the test logs?"

yes

> "Minor nitpick, but is it necessary to change the notebooks in this PR? I anticipate that there might be some merge conflicts with https://github.com/fmfn/BayesianOptimization/pull/419."

Fair point - it wasn't necessary, but I'm going to just merge for now as it's less work :-)

PR #429, opened 2023-05-24
Fixes issue-428: TargetSpace.max() bug in constrained space

### LINKED ISSUE #428: Constrained Target Space occasionally outputs the wrong parameters for a maximum
**Describe the bug**
If two points are registered with the same maximum target value, but the first one is disallowed and the second is not, then `.max()` outputs the disallowed one.

**To Reproduce**
A concise, self-contained code snippet that reproduces the bug you would like to report.
```python

import pytest
import numpy as np
from bayes_opt.constraint import ConstraintModel
from bayes_opt.target_space import TargetSpace
from bayes_opt.util import NotUniqueError


def target_func(**kwargs):
    # arbitrary target func
    return sum(kwargs.values())


PBOUNDS = {'p1': (0, 1), 'p2': (1, 100)}

def test_max_with_constraint():
    constraint = ConstraintModel(lambda p1, p2: p1-p2, -2, 2)
    space = TargetSpace(target_func, PBOUNDS, constraint=constraint)

    assert space.max() == None
    space.probe(params={"p1": 1, "p2": 2}) # Feasible
    space.probe(params={"p1": 0, "p2": 5}) # Unfeasible
    space.probe(params={"p1": 5, "p2": 8}) # Unfeasible
    space.probe(params={"p1": 2, "p2": 3}) # Feasible
    space.probe(params={"p1": 1, "p2": 6}) # Unfeasible
    assert space.max() == {"params": {"p1": 2, "p2": 3}, "target": 5, "constraint": -1}
```

**Additional context**
This seems to happen because [this line](https://github.com/bayesian-optimization/BayesianOptimization/blob/0d5105dcc25263b2911262dafcb61da28209ac8f/bayes_opt/target_space.py#L293) simply takes the first parameter combo with the matching target value, but does not check whether this is in fact the allowed one.
CC: @leandrobbraga

leandrobbraga:
This commit fixes the bug in TargetSpace.max() when the space is constrained which lead to sometimes returning the parameters from a unfeasible point when its target value coincided with the best feasible point.

Closes #428

### DIFF
--- a/bayes_opt/target_space.py
+++ b/bayes_opt/target_space.py
@@ -43,7 +43,7 @@ class TargetSpace(object):
             If True, the optimizer will allow duplicate points to be registered.
             This behavior may be desired in high noise situations where repeatedly probing
             the same point will give different answers. In other situations, the acquisition
-            may occasionaly generate a duplicate point.
+            may occasionally generate a duplicate point.
         """
         self.random_state = ensure_rng(random_state)
         self._allow_duplicate_points = allow_duplicate_points
@@ -70,7 +70,7 @@ class TargetSpace(object):
         self._constraint = constraint
 
         if constraint is not None:
-            # preallocated memory for constraint fulfillement
+            # preallocated memory for constraint fulfillment
             if constraint.lb.size == 1:
                 self._constraint_values = np.empty(shape=(0), dtype=float)
             else:
@@ -170,7 +170,7 @@ class TargetSpace(object):
 
         Notes
         -----
-        runs in ammortized constant time
+        runs in amortized constant time
 
         Example
         -------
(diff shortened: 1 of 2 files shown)


till-m:
Thanks! Better be careful @leandrobbraga, if you do much more I'll have to add you as maintainer.

till-m:
@leandrobbraga jokes aside, would you be interested in becoming a maintainer on this project?
@fmfn @bwheelz36 I think we could use some more help, right?

bwheelz36:
100%

leandrobbraga:
Thank you, currently I don't have much bandwidth to be a full maintainer, but sure I can help here and there with an issue 🙂

Issue #434, opened 2023-07-05
Replace custom colour implementation with colorama colours


till-m:
Currently, this package implements its own colours. Since colorama is already a dependency to fix the display of said colours on windows machines, we may aswell use colorama's colour implementation and reduce the codebase a bit.

I don't see any drawbacks to this except the additional effort it takes to refactor the codebase for this change.

**Are you able and willing to implement this feature yourself and open a pull request?**
- [x] Yes, I can provide this feature.

till-m:
done #457

Issue #427, opened 2023-05-16
settle on a docstring format and update all docstrings


bwheelz36:
As discussed in #425.

Suggest using the numpy format, and integrating [pydocstyle](https://numpydoc.readthedocs.io/en/latest/) into test framework

TODO:
- [x] Update `constraint.py` #440 
- [x] Update `logger.py` #435 
- [x] Update `util.py` #435 
- [x] Update `domain_reduction.py` #441 
- [x] Update `observer.py` #445 
- [x] Update `bayesian_optimization.py`#445
- [x] Update `event.py` #445 
- [x] Update `target_space.py` #440 
- [x] Integrate `pydocstyle` into testing. #445 
- [x] Use `numpydoc`

bwheelz36:
this will probably be helpful too:
https://github.com/dadadel/pyment

till-m:
TODO:
- [ ] Update `constraint.py`
- [ ] Update `logger.py`
- [ ] Update `util.py`
- [ ] Update `domain_reduction.py`
- [ ] Update `observer.py`
- [ ] Update `bayesian_optimization.py`
- [ ] Update `event.py`
- [ ] Update `target_space.py`
- [ ] Integrate `pydocstyle` into testing.

till-m:
I have started doing some work on this. I also ran pylint over the codebase in parallel and started fixing some of code smells it reports. Since this is a rather large undertaking I would suggest proceeding in the following way:

- We fork the main branch to a separate branch in this project called something along the lines of `docstring-overhaul` or similar.
- We successively process the docstrings of each file and after finishing 1-3 files open a pull request to the `docstring-overhaul` branch
- After all files have been processed, we merge the `docstring-overhaul` branch into master.

This will make the entire thing more granular and not result in a 4000 lines PR that needs to be reviewed. Any thoughts or objections?

bwheelz36:
Great work. 
Honestly, not sure how much I will be able to help. I'm pretty busy with my new job. However the process sounds good :-)

PR #435, opened 2023-07-10
Replace custom colour implementation, add docs for `logger.py`, `util.py`

till-m:
cf. #427 #434

Also contains fixes for some code smells reported by `pylint`.

NB: This PR is not against master, but against a separate branch.


till-m:
@leandrobbraga a review of yours would also be appreciated, if you have the time.

bayes_opt/logger.py
```diff
@@ -117,14 +212,25 @@ def _is_new_max(self, instance):
         return instance.max["target"] > self._previous_max
 
     def update(self, event, instance):
+        """Handle incoming events.
+
+        Parameters
+        ----------
+        event : str
```
leandrobbraga, line 219: Is this a string or an `Events` object? By looking at the code it is comparing `event == Events.OPTIMIZATION_START`, which implies that it is an Events object.

bayes_opt/logger.py
```diff
@@ -135,18 +241,43 @@ def update(self, event, instance):
...
+            One of the values associated with `Events.OPTIMIZATION_START`,
+            `Events.OPTIMIZATION_STEP` or `Events.OPTIMIZATION_END`.
+            
+        instance : bayesian_optimization.BayesianOptimization
+            The instance associated with the step.
+
+        """
         if event == Events.OPTIMIZATION_STEP:
```
leandrobbraga, line 281: Same thing here, it seems that we expect `event` to be of `Events` type

bayes_opt/logger.py
```diff
@@ -1,82 +1,155 @@
...
+        Returns
+        -------
+        A stringified, formatted version of `x`.
+        """
         if 5 > self._default_cell_size:
-            if x == True:
+            if x:
                 x_ = 'T'
-            elif x == False:
+            else:
```
leandrobbraga, line 114: These two codes have different behavior, it seems that the old one was intently done with if / elif.

```
>>> x = "false"
>>> if x:
...    print("got 'true'!")
...
got 'true'!
```

In the previous code it would ignore anything that's not a boolean

leandrobbraga (approved):

bayes_opt/logger.py
till-m: These `Events.OPTIMIZATION_START` etc. are properties of the events classed with assigned string values `optimization:start` etc. I thought about going for `Event` as type, but that is technically not correct. Maybe it would be good to define `Events` as a proper [Enum](https://docs.python.org/3/library/enum.html) but I wouldn't do that in this PR.

bayes_opt/logger.py
till-m: > it seems that the old one was intently done with if / elif

I think you're giving me too much credit here 🙃 

I do see how this could potentially print something wrong if called with the wrong argument type. What do you think would be the best way to handle this case?

PR #398, opened 2023-02-26
Issues/394

LemurPwned:
Hey,
here's my suggestion to fix the trimming issue. The tests seem to pass, and the global bounds should be respected.

A suggestion perhaps to add contribution guideline (formatting etc.). 
Sorry for the initial erroneous commit -- please squash if it's ok.

Issue #394, opened 2023-01-25
Sequential Domain Reduction does not respect global bounds


raghavchandra123:
**Describe the bug**
the minimum window width is imposed after the global bounds trim rather than before which I think means that if a minimum window is wide enough then it can escape global bounds. This applies to the trim method of the SequentialDomainReductionTransformer.

**To Reproduce**
For example:
`def _trim(new_bounds: np.array, global_bounds: np.array) -> np.array:
	min_window = 1
        for i, variable in enumerate(new_bounds):
            if variable[0] < global_bounds[i, 0]:
                variable[0] = global_bounds[i, 0]
            if variable[1] > global_bounds[i, 1]:
                variable[1] = global_bounds[i, 1]
        for i, entry in enumerate(new_bounds):
            if entry[0] > entry[1]:
                new_bounds[i, 0] = entry[1]
                new_bounds[i, 1] = entry[0]
            window_width = abs(entry[0] - entry[1])
            if window_width < 1:
                new_bounds[i, 0] -= (min_window - window_width) / 2.0
                new_bounds[i, 1] += (min_window - window_width) / 2.0

        return new_bounds
new_bounds = np.concatenate((np.ones((5,1))*0.1,np.ones((5,1))),axis=1)
global_bounds = np.concatenate((np.ones((5,1))*-1,np.ones((5,1))),axis=1)*0.5
print(new_bounds)
print(global_bounds)
print(_trim(new_bounds,global_bounds))`
prints:
 [[0.1 1. ]
 [0.1 1. ]
 [0.1 1. ]
 [0.1 1. ]
 [0.1 1. ]]
[[-0.5  0.5]
 [-0.5  0.5]
 [-0.5  0.5]
 [-0.5  0.5]
 [-0.5  0.5]]
[[-0.2  0.8]
 [-0.2  0.8]
 [-0.2  0.8]
 [-0.2  0.8]
 [-0.2  0.8]]

where the first array is new_bounds before trim, second is global bounds and third is the trimmed version which is outside global bounds

**Expected behavior**
I think the global bounds check and minimum window should be combined in a way to respect both, e.g. for the example above it should return the new_bounds as -0.5 and 0.5. 

**Screenshots**

till-m:
Hey @raghavchandra123,

thanks for reporting this. Would you be willing to submit a PR to make this change? I would be happy to review it!

till-m:
@bwheelz36 I think this can be closed now :)

PR #441, opened 2023-08-14
bounds_transformer could bypass global_bounds due to the test logic within _trim function in domain_reduction.py

perezed00:
Previously, when the upper limit of  new_bounds < lower limit of global_bounds, the new_bounds could bypass the global_bounds.  The proposed changes requires the little change to the existing code, but could cause one of the entries in new_bounds to stagnate at the global_bounds if the new_bounds entry is always outside of the global_bounds.

### DIFF
--- a/bayes_opt/domain_reduction.py
+++ b/bayes_opt/domain_reduction.py
@@ -37,7 +37,10 @@ class SequentialDomainReductionTransformer(DomainTransformer):
         self.minimum_window_value = minimum_window
 
     def initialize(self, target_space: TargetSpace) -> None:
-        """Initialize all of the parameters"""
+        """Initialize all of the parameters.
+        """
+    
+        # Set the original bounds
         self.original_bounds = np.copy(target_space.bounds)
         self.bounds = [self.original_bounds]
 
@@ -48,6 +51,7 @@ class SequentialDomainReductionTransformer(DomainTransformer):
         else:
             self.minimum_window = [self.minimum_window_value] * len(target_space.bounds)
 
+        # Set initial values
         self.previous_optimal = np.mean(target_space.bounds, axis=1)
         self.current_optimal = np.mean(target_space.bounds, axis=1)
         self.r = target_space.bounds[:, 1] - target_space.bounds[:, 0]
@@ -73,14 +77,13 @@ class SequentialDomainReductionTransformer(DomainTransformer):
         self._window_bounds_compatibility(self.original_bounds)
 
     def _update(self, target_space: TargetSpace) -> None:
-
+        """ Updates contraction rate, window size, and window center.
+        """
         # setting the previous
         self.previous_optimal = self.current_optimal
         self.previous_d = self.current_d
-
-        self.current_optimal = target_space.params[
-            np.argmax(target_space.target)
-        ]
+        
+        self.current_optimal = self._windowed_max(target_space)
 
         self.current_d = 2.0 * (self.current_optimal -
                                 self.previous_optimal) / self.r
@@ -96,46 +99,101 @@ class SequentialDomainReductionTransformer(DomainTransformer):
             np.abs(self.current_d) * (self.gamma - self.eta)
 
         self.r = self.contraction_rate * self.r
+        
+    def _windowed_max(self, target_space: TargetSpace) -> np.array:
+        """Returns the parameters that produce the greatest target value within its bounds.
+        """
+        # extract the components we need from the target space
+        params = np.copy(target_space.params.T)
+        target = np.copy(target_space.target)
+        bounds = np.copy(target_space.bounds)
+        
+        # create a mask by checking each params against its bounds
+        mask = np.zeros_like(target)
+        for n, row in enumerate(params):
+            lower_bound = bounds[n, 0]
+            upper_bound = bounds[n, 1]
+            mask += (row <= lower_bound).astype(int)
+            mask += (row >= upper_bound).astype(int)
+        mask = mask<1
+
+        # apply the mask
+        params = params.T[mask]
+        targets = target[mask]
+
+        best_params = params[np.argmax(targets)]
+
+        return best_params
 
     def _trim(self, new_bounds: np.array, global_bounds: np.array) -> np.array:
-        """ Tests new_bounds to enforce global_bounds and minimum_window."""
-        # sort the order of the bounds for each parameter
+        """
+        Adjust the new_bounds and verify that they adhere to global_bounds and minimum_window.
+
+        Parameters:
+        -----------
+        new_bounds : np.array
+            The proposed new_bounds that (may) need adjustment.
+
+        global_bounds : np.array
+            The maximum allowable bounds for each parameter.
+
+        Returns:
+        --------
+        new_bounds : np.array
+            The adjusted bounds after enforcing constraints.
+        """
+
+        #sort bounds
         new_bounds = np.sort(new_bounds)
 
-        # check if the new bounds exceed the global bounds
+        # Validate each parameter's bounds against the global_bounds
         for i, pbounds in enumerate(new_bounds):
-            # check if lower bound exceeds range 
+            # If the lower bound of the parameter is outside the global bounds, reset the lower bound
             if (pbounds[0] < global_bounds[i, 0] or pbounds[0] > global_bounds[i, 1]):
                 pbounds[0] = global_bounds[i, 0]
-                warn("""Domain reduction warning:
-                    A parameter's lower bound has exceeded its global boundaries.
-                    The offensive boundary has been reset, but the optimizer may not converge.""")
-                
-            # check if upper bound exceeds range
+                warn("""Domain Reduction Warning:
+                    A parameter's lower bound has exceeded its global limit.
+                    The offensive boundary has been reset, but be cautious of optimizer convergence.""")
+
+            # If the upper bound bound of the parameter is outside the global bounds, reset the lower bound
             if (pbounds[1] > global_bounds[i, 1] or pbounds[1] < global_bounds[i, 0]):
                 pbounds[1] = global_bounds[i, 1]
                 warn("""Domain reduction warning:
-                    A parameter's upper bound has exceeded its global boundaries.
-                    The offensive boundary has been reset, but the optimizer may not converge.""")
-     
-        for i, entry in enumerate(new_bounds):
-            window_width = abs(entry[0] - entry[1])
-            if window_width < self.minimum_window[i]:
-                dw = (self.minimum_window[i] - window_width) / 2.0
-                left_expansion_space = abs(global_bounds[i, 0] - entry[0]) # should be non-positive
-                right_expansion_space = abs(global_bounds[i, 1] - entry[1]) # should be non-negative
-                # conservative
-                dw_l = min(dw, left_expansion_space)
-                dw_r = min(dw, right_expansion_space)
-                # this crawls towards the edge
-                ddw_r = dw_r + max(dw - dw_l, 0)
-                ddw_l = dw_l + max(dw - dw_r, 0)
-                new_bounds[i, 0] -= ddw_l
-                new_bounds[i, 1] += ddw_r
+                    A parameter's lower bound has exceeded its global limit.
+                    The offensive boundary has been reset, but be cautious of optimizer convergence.""")
+
+        # Adjust new_bounds to ensure they respect the minimum window width for each parameter
+        for i, pbounds in enumerate(new_bounds):
+            current_window_width = abs(pbounds[0] - pbounds[1])
+
+            # If the window width is less than the minimum allowable width, adjust it
+            # Note that minimum_window < width of the global bounds one side always has more space than required
+            if current_window_width < self.minimum_window[i]:
+                width_deficit = (self.minimum_window[i] - current_window_width) / 2.0
+                available_left_space = abs(global_bounds[i, 0] - pbounds[0])
+                available_right_space = abs(global_bounds[i, 1] - pbounds[1])
+                
+                # determine how much to expand on the left and right
+                expand_left = min(width_deficit, available_left_space)
+                expand_right = min(width_deficit, available_right_space)
+                
+                # calculate the deficit on each side
+                expand_left_deficit = width_deficit - expand_left
+                expand_right_deficit = width_deficit - expand_right
+
+                # shift the deficit to the side with more space
+                adjust_left = expand_left + max(expand_right_deficit, 0)
+                adjust_right = expand_right + max(expand_left_deficit, 0)
+                
+                # adjust the bounds
+                pbounds[0] -= adjust_left
+                pbounds[1] += adjust_right
+
         return new_bounds
 
     def _window_bounds_compatibility(self, global_bounds: np.array) -> bool:
-        """Checks if global bounds are compatible with the minimum window sizes."""
+        """Checks if global bounds are compatible with the minimum window sizes.
+        """
         for i, entry in enumerate(global_bounds):
             global_window_width = abs(entry[1] - entry[0])
             if global_window_width < self.minimum_window[i]:
@@ -146,7 +204,8 @@ class SequentialDomainReductionTransformer(DomainTransformer):
         return {param: bounds[i, :] for i, param in enumerate(parameters)}
 
     def transform(self, target_space: TargetSpace) -> dict:
-
+        """Reduces the bounds of the target space.
+        """
         self._update(target_space)
 
         new_bounds = np.array(
@@ -156,6 +215,6 @@ class SequentialDomainReductionTransformer(DomainTransformer):
             ]
         ).T
 
-        self._trim(new_bounds, self.original_bounds)
+        new_bounds = self._trim(new_bounds, self.original_bounds)
         self.bounds.append(new_bounds)
         return self._create_bounds(target_space.keys, new_bounds)
--- a/bayes_opt/domain_reduction.py
+++ b/bayes_opt/domain_reduction.py
@@ -83,7 +83,7 @@ class SequentialDomainReductionTransformer(DomainTransformer):
         self.previous_optimal = self.current_optimal
         self.previous_d = self.current_d
         
-        self.current_optimal = self._windowed_max(target_space)
+        self.current_optimal = target_space.params_to_array(target_space.max()['params'])
 
         self.current_d = 2.0 * (self.current_optimal -
                                 self.previous_optimal) / self.r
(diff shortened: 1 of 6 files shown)


till-m:
This should've been fixed in #398 (c.f. also #394). I don't think this change has been deployed to PyPI yet though.

perezed00:
Thanks for linking me to those posts, I was not aware of them. However, this issue is distinct from  #394, which addresses a window-width bypass. In `1.4.3` the global_bounds can still be bypassed when the suggested upper bound is smaller than the original lower bound. This PR addresses the current issue by identifying if either of the suggested new_bounds are outside of the original_bounds, in which case it sets the suggested new_bound value to the global_bound value. 

This issue arises, for instance, when an optimizer instance loads a log that was generated by an instance with bounds greater than those of the current instance. 

I will return to this PR to add a test case to `tests/test_seq_domain_red.py` and to verify the order of the various checks in `_trim`. I suspect that the proposed solution would be more robust if the entries of new_bounds are sorted before the global bounds are enforced.

till-m:
Hey @perezed00,

just to confirm, the case you're trying to cover is that sometimes, the lower bound can be greater than the upper bound (i.e. the bounds are out of order) and then the lower bound is not trimmed properly and allowed to be greater than the global bound?

perezed00:
No, the case I've identified is roughly as follows:
Suppose the global bounds for a variable are [5, 10]. If `bounds_transformer` suggests new_bounds = [1, 4], then _both_ bounds are smaller than the lower global_bound. `v1.4.3` only checks if _one_ of the bounds is smaller than global limit, so the second bound sneaks in.

Suppose the global_bounds for a variable are [x_lower, x_upper]. 
Suppose bounds_transformer suggests new_bounds=[nx_lower, nx_upper]. Assume nx_lower<nx_upper.

If nx_lower < x_lower, `v1.4.3` catches the error.
If (nx_lower< x_lower  and  nx_upper < x_lower), `v1.4.3` does not catch the error. (Caught by this PR)
If (nx_lower > nx_upper  and nx_upper < x_lower), needs to be checked.

till-m:
Thanks, that makes sense! I guess some more testing is in order. If the case you describe only shows up when loading logs from a differently configured optimizer, it makes sense that it wasn't covered before.
I assume you now have a solid understanding of the trim logic. Would you mind adding some comments to the code as part of this PR to make it clear what's happening in the `_trim` function?

perezed00:
Sounds good. I don't mind adding some comments to the `_trim` function. 
The case I've described can also arise after the use of `optimizer.set_bounds(new_bounds={})` 

In summary, I'll revise this PR after doing the following:
- Add comments to `_trim`
- Verify the (nx_lower > nx_upper and nx_upper < x_lower) case described above
- Check if the entries of new_bounds should be sorted before global bound enforcement
- Add a test case to `tests/test_seq_domain_red.py`

It will probably be a few days before I get a chance to implement these, but I'll post again when complete.

till-m:
Hey @perezed00, did you get around to have another look at this? Seems like the issue might've popped up for another user so I'd be eager to get this fixed.

perezed00:
Hi, @till-m 

I have updated this pull request. Regarding the list of tasks above, here's what was done:
- Comments were added to `_trim`. The section of `_trim` that deals with the window size could still use more comments, however.
- The (nx_lower > nx_upper and nx_upper < x_lower) case described above is tested, and not and issue
- The entries of new bounds are now sorted before global bounds are enforced. This change was not **needed** but it makes the organization of `_trim` a little easier to follow
- Test cases were added to `tests/test_seq_domain_red.py`

Additionally, the name of variables in `_trim` was made more uniform. With the exception of `test_notebooks_run.py` all of the tests in `/tests` pass. `test_notebooks_run.py` was not tested.

bayes_opt/domain_reduction.py
```diff
@@ -97,15 +97,23 @@ def _update(self, target_space: TargetSpace) -> None:
...
-            if variable[0] < global_bounds[i, 0]:
-                variable[0] = global_bounds[i, 0]
-            if variable[1] > global_bounds[i, 1]:
-                variable[1] = global_bounds[i, 1]
+        """ Tests new_bounds to enforce global_bounds and minimum_window."""
+        # sort the order of the bounds for each parameter
+        for i, pbounds in enumerate(new_bounds):
+            if pbounds[0] > pbounds[1]:
+                new_bounds[i, 0] = pbounds[1]
+                new_bounds[i, 1] = pbounds[0]
```
till-m, line 105: I think this can malfunction since iterating over an array (`__next__`) returns a pointer to the original object's data.

```python
import numpy as np
new_bounds = np.array([[2., 1.], [4., 3.]])
for i, pbounds in enumerate(new_bounds):
    if pbounds[0] > pbounds[1]:
        new_bounds[i, 0] = pbounds[1]
        new_bounds[i, 1] = pbounds[0]
```
Out:
```
array([[1., 1.],
       [3., 3.]])
```

bayes_opt/domain_reduction.py
```diff
@@ -97,15 +97,23 @@ def _update(self, target_space: TargetSpace) -> None:
...
+
+        # check if the new bounds exceed the global bounds
+        for i, pbounds in enumerate(new_bounds):
+            # check if lower bound exceeds range 
+            if (pbounds[0] < global_bounds[i, 0] or pbounds[0] > global_bounds[i, 1]):
+                pbounds[0] = global_bounds[i, 0]
+            # check if upper bound exceeds range
+            if (pbounds[1] > global_bounds[i, 1] or pbounds[1] < global_bounds[i, 0]):
+                pbounds[1] = global_bounds[i, 1]
```
till-m, line 114: This can essentially cause a reset to the initial global bounds if we pan outside them completely.
1. Can this case occur naturally, or only if the user manually messes with the bounds?
2. If it can happen naturally, can the bounds transformer recover from this and resume normal operation?

till-m (commented):
Looks good to me in general. Some small questions and things to verify.

bayes_opt/domain_reduction.py
perezed00: I had a suspicion this could be a problem, but that's how it was previously implemented, so I left it as-is. Anyways, this can be handled by np.sort(pbounds) instead. I'll make the change.

bayes_opt/domain_reduction.py
perezed00: FYI, I think this possible route to non-convergence might already be present in `v1.4.3`

bayes_opt/domain_reduction.py
till-m: Yeah, you're right, it must've slipped by the last time. Thanks for handling it!

till-m:
Regarding your alternative solution -- I like your suggestion, maybe the appropriate way to implement this would be to restrict `TargetSpace.max()` to return the maximum obeying the global bounds. The problem is, that the global bounds are currently only stored in the transformer, not in the target space. (cc: @fmfn @bwheelz36 @leandrobbraga)

Also, I think it'd be good to have a warning to inform the user that the transformer observed a maximum outside the global bounds and why this can happen.

bayes_opt/domain_reduction.py
perezed00: done

perezed00:
Since we have identified a few separate issues, I propose we implement the solutions in separate PRs. 

Here is how I see the issues:
`Issue #0`: The transformer is suggesting boundaries that exceed the global bounds. 
`Issue #1`: Some erroneous boundaries do not get caught by existing `_trim` logic.
`Issue #2`: Current solutions to issue `#1` can prevent the optimizer from converging in certain cases. 

`Issue #0 `is the root cause of issues `#1` and `#2`.  This PR addresses `#1`, but leaves `#0` and `#2` unresolved. The [suggested ](https://github.com/bayesian-optimization/BayesianOptimization/pull/441#issuecomment-1707115417)solution above may address `#0` directly. For the sake of organized updates lets keep this discussion focused on `#1` and make a new PR for `#0`.  What do you think? If you agree I can make a new branch / PR.

I don't mind if this PR get's closed without integration (I'll leave that decision to you), I just don't want this single PR to become a discussion that is too long to follow.

till-m:
Hi @perezed00,

my apologies, for not getting back to you, I have been rather busy recently.
I think it is generally better to tackle issues at the root. How would your solution to issue #0 look like (mathematically speaking)? I think there are several ways of doing what you're trying to accomplish with subtle, but important differences.

perezed00:
Hi @till-m,

No problem, I have also been very busy. Basically, the if the transformer sees a set of parameters in the target_space that optimize the function, then it tries to converge to that set of parameters. Mathematically, can mask the target_space so that the transformer only sees the parameters within the global bounds.  I implemented a windowing function within `domain_reduction.py' that doesn't require modifying other portions of the code base. I'll upload that now, along with some improved comments. 

During this process, I learned that the mask will not stop the transformer from suggesting bounds outside of the `global_bounds`, since that process also depends on the zoom/pan parameters. I am not sure that it can be avoided, and I don't think it needs to be addressed. This is because the `_trim` function enforces the global bounds.

On the other hand, as far as I can tell, the mask does drastically improve convergence. I believe that a transformer with a mask is as good as (or better than) a "normal" transformer at converging.  This can (should?) be explicitly tested.

If, in your opinion, the convergence seems okay, we might want to suppress the convergence warning since it is quite common at this point.

bayes_opt/domain_reduction.py
```diff
@@ -96,46 +99,101 @@ def _update(self, target_space: TargetSpace) -> None:
             np.abs(self.current_d) * (self.gamma - self.eta)
 
         self.r = self.contraction_rate * self.r
+        
+    def _windowed_max(self, target_space: TargetSpace) -> np.array:
```
till-m, line 103: I think it would be better to ensure that `target_space.max()` respects the pbounds and then use that function to get the maximum. Otherwise we will have two functions doing pretty much the same thing, which adds maintenance overhead (and the SDRT should really respect constraints, too).

bayes_opt/domain_reduction.py
```diff
@@ -96,46 +99,101 @@ def _update(self, target_space: TargetSpace) -> None:
...
             if (pbounds[0] < global_bounds[i, 0] or pbounds[0] > global_bounds[i, 1]):
                 pbounds[0] = global_bounds[i, 0]
-                warn("""Domain reduction warning:
-                    A parameter's lower bound has exceeded its global boundaries.
-                    The offensive boundary has been reset, but the optimizer may not converge.""")
-                
-            # check if upper bound exceeds range
+                warn("""Domain Reduction Warning:
+                    A parameter's lower bound has exceeded its global limit.
+                    The offensive boundary has been reset, but be cautious of optimizer convergence.""")
```
till-m, line 156: `optimizer` generally refers to the bayesian optimizer, which should still converge to something, just slower compared to when the SDRT is used. This message might be confusing for users who could think that the entire optimization procedure has broken down. You could write, e.g.:

```suggestion
                    The offensive boundary has been reset, but the optimizer may proceed without domain reduction.""")
```
but not super attached to it and there's probably a better way to phrase that.

bayes_opt/domain_reduction.py
```diff
@@ -96,46 +99,101 @@ def _update(self, target_space: TargetSpace) -> None:
...
-                # conservative
-                dw_l = min(dw, left_expansion_space)
-                dw_r = min(dw, right_expansion_space)
-                # this crawls towards the edge
-                ddw_r = dw_r + max(dw - dw_l, 0)
-                ddw_l = dw_l + max(dw - dw_r, 0)
-                new_bounds[i, 0] -= ddw_l
-                new_bounds[i, 1] += ddw_r
+                    A parameter's lower bound has exceeded its global limit.
+                    The offensive boundary has been reset, but be cautious of optimizer convergence.""")
```
till-m, line 163: as above. Note that both of them state `lower bound has exceeded its global limit` but one of the cases presumably concerns the upper bound.

bayes_opt/domain_reduction.py
```diff
@@ -96,46 +99,101 @@ def _update(self, target_space: TargetSpace) -> None:
...
+                expand_left_deficit = width_deficit - expand_left
+                expand_right_deficit = width_deficit - expand_right
+
+                # shift the deficit to the side with more space
+                adjust_left = expand_left + max(expand_right_deficit, 0)
+                adjust_right = expand_right + max(expand_left_deficit, 0)
+                
+                # adjust the bounds
+                pbounds[0] -= adjust_left
+                pbounds[1] += adjust_right
```
till-m, line 190: I quite like this code and the variable names are lovely :)

till-m (changes requested):
Sorry for the delay...
Biggest thing from my side is the fact that the `max_window` should be calculated using the target space. Rest looks good to me :)

perezed00:
Sorry for the late reply here , I am in the final stretch of dissertation work. I will finish this in 5 days.

till-m:
No worries. Good luck with the dissertation!

till-m:
Hey @perezed00, don't mean to rush you, but any update on this? :)

bayes_opt/domain_reduction.py
perezed00: This has been implemented. Thanks for the suggestion. I like it better

bayes_opt/domain_reduction.py
perezed00: I rewrote when the warnings were triggered to be more specific and reduce excessive warnings. Please see updated version.

perezed00:
Hey @till-m ,

Thanks for your patience. The changes have been implemented. target_space.max() now respects the bounds and I improved the warning messages. Warnings are now more specific and less frequent. Let me know if you see any other issues.

Cheers

tests/test_seq_domain_red.py
```diff
@@ -171,13 +171,20 @@ def verify_bounds_in_range(new_bounds, global_bounds):
...
     new_bounds = np.array( [[-20, -50], [-10, 10]] )
     trimmed_bounds = bounds_transformer._trim(new_bounds, global_bounds)
     assert verify_bounds_in_range(trimmed_bounds, global_bounds)
+
+if __name__ == '__main__':
+    r"""
+    CommandLine:
+        python tests/test_seq_domain_red.py
+    """
+    pytest.main([__file__])
```
perezed00, line 190: this is consistent with the other test files.

tests/test_target_space.py
```diff
@@ -193,6 +202,7 @@ def test_max():
     assert space.max() == {"params": {"p1": 5, "p2": 4}, "target": 9}
 
 def test_max_with_constraint():
+    PBOUNDS = {'p1': (0, 10), 'p2': (1, 100)}
```
perezed00, line 205: added local pbounds since the `.target_max()` method would now throw an error.

till-m:
hey, I'm a bit swamped right now, will have a look as soon as I can. Thanks for the contributions :)

tests/test_seq_domain_red.py
till-m: I think these should be removed, as they are not needed anymore, but it's good that you made sure it's consistent. Let's keep it like this for now, and we can delete them all in the future.

tests/test_target_space.py
```diff
@@ -166,10 +166,10 @@ def test_random_sample():
 def test_y_max():
     space = TargetSpace(target_func, PBOUNDS)
     assert space._target_max() == None
-    space.probe(params={"p1": 1, "p2": 2})
-    space.probe(params={"p1": 5, "p2": 1})
```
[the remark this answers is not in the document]
till-m: yeah, I think this works as intended. I saw that the case where there is a better point outside of pbounds is also covered, so I think this looks good 👍

tests/test_target_space.py
```diff
@@ -180,11 +180,20 @@ def test_y_max_with_constraint():
...
+    space.probe(params={"p1": 0, "p2": 1})
+    assert space._target_max() == 3
+
 
 
 def test_max():
+    PBOUNDS = {'p1': (0, 10), 'p2': (1, 100)}
     space = TargetSpace(target_func, PBOUNDS)
-
+
```
till-m, line 196: nitpick: indentation of empty line

bayes_opt/target_space.py
```diff
@@ -263,6 +279,14 @@ def random_sample(self):
             data.T[col] = self.random_state.uniform(lower, upper, size=1)
         return data.ravel()
 
+    def _pbound_mask(self):
```
till-m: Is this function unused? If no, what's its purpose?

tests/test_target_space.py
```diff
@@ -180,11 +180,20 @@ def test_y_max_with_constraint():
...
     assert space._target_max() == 3
 
+def test_y_max_within_pbounds():
+    space = TargetSpace(target_func, PBOUNDS)
+    assert space._target_max() == None
+    space.probe(params={"p1": 1, "p2": 2})
+    space.probe(params={"p1": 5, "p2": 1})
+    space.probe(params={"p1": 0, "p2": 1})
+    assert space._target_max() == 3
+
```
till-m, line 190: nitpick: One newline too many
```suggestion

```

bayes_opt/target_space.py
```diff
@@ -116,6 +116,22 @@ def constraint_values(self):
...
+    def mask(self):
+        mask = np.ones_like(self.target, dtype=bool)
+
+        # mask points that don't satisfy the constraint
+        if self._constraint is not None:
+            mask &= self._constraint.allowed(self._constraint_values)
+
+        # mask points that are outside the bounds
+        if self._bounds is not None:
+            within_bounds = np.all((self._bounds[:, 0] <= self._params) &
```
till-m, line 132: @perezed00 @bwheelz36 @leandrobbraga @fmfn 
What do you think, does it make sense trigger warning here if we find values outside pbounds?

till-m (approved):
Overall I think this is basically ready for merge, excluding the 3 minor comments + the warning discussion.
@osullivryan If you have the time, I think a review from your side would also be appreciated.

till-m:
Thanks for adding so much high-quality documentation, by the way!

bayes_opt/target_space.py
till-m: Whoops, you already removed this. My bad!

osullivryan:
Hey @perezed00  and @till-m , 

I took a look. Nice catch and fix! Awesome!

bayes_opt/target_space.py
perezed00: There are three ways for parameters to exist beyond the pbounds:
1. Existing pbounds are shrunk using `target_space.set_bounds(small_pbounds)`
2. They are manually (i.e., intentionally) added using `target_space.register(x, y)`
3. Logs are loaded that don't respect the pbounds. 

I don't think that #1 should produce a warning since this intended behavior. To tackle #2 and #3 without affecting #1, we could issue a warning when a point outside the bounds is **registered**, rather than found?

tests/test_seq_domain_red.py
perezed00: I found them helpful when writing the tests, since I could just "run" the test files, but I understand if you decide to remove them.

bayes_opt/target_space.py
till-m: That sounds good, would you mind implementing the warning + adding a test for it?

tests/test_seq_domain_red.py
till-m: let's keep them for now.

bayes_opt/target_space.py
perezed00: This warning has been added and all warnings have been slightly updated for improved formatting.

till-m:
From my side, I think this is good to go. @bwheelz36 @leandrobbraga @fmfn feel free to review. If you have no comments, I will merge this PR soon.

@perezed00 thank you very much for your contribution. As you may have noticed, we are a bit short on maintainers for this project currently. Based on the way you wrote this PR, the thoughtfulness of your approaches and the clarity of the code you wrote, I think it would be very valuable to have you on the team. If you're interested, please let me know :)

perezed00:
@till-m Thanks for the invite to join the team. I accept, I would be happy to join. Just a bit of a warning however: I defend my dissertation in April, so I might be less responsive a couple months before then.

Cheers!

bwheelz36:
Looks good to me. 
regarding the codecov checks, I'm not sure if you knew but you can check the code cov analysis of this PR here:
https://app.codecov.io/gh/bayesian-optimization/BayesianOptimization/pull/441
Apparently there are a few lines like `if (pbounds[0] > global_bounds[i, 1]):` which aren't being tested - should the test case be covering these?

PR #444, opened 2023-10-04
Update domain_reduction.py

lm314:
Adds ability to pass a dictionary as the minimum_window argument in SequentialDomainReductionTransformer. Deals with issue where TargetSpace sorts the pBounds dictionary, causing the order of the list used for minimum_window to possibly not match the order of the stored boundaries.

### DIFF
--- a/tests/test_seq_domain_red.py
+++ b/tests/test_seq_domain_red.py
@@ -68,6 +68,7 @@ def test_bound_x_maximize():
     assert not (standard_optimizer._space.bounds ==
                 mutated_optimizer._space.bounds).any()
 
+
 def test_minimum_window_is_kept():
     bounds_transformer = SequentialDomainReductionTransformer(minimum_window=1.0)
     pbounds = {'x': (-0.5, 0.5), 'y': (-1.0, 0.0)}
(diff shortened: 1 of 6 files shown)


tests/test_seq_domain_red.py
```diff
@@ -143,4 +143,21 @@ def test_exceeded_bounds():
...
\ No newline at end of file
+            )
+
+def test_minimum_window_dict_ordering():
+    """Tests if dictionary input for minimum_window is reordered the same as pbounds"""
+    window_ranges = {'y': 1, 'x': 3,'w': 1e5}
+    bounds_transformer = SequentialDomainReductionTransformer(minimum_window=window_ranges)
+    pbounds = {'y': (-1, 1),'w':(-1e6,1e6), 'x': (-10, 10)}
+    
+    try:
```
till-m, line 154: The check shouldn't be wrapped in a `try-except` statement.

till-m (changes requested):
LGTM, except for the minor change in the test. I would however wait with merging this until #441 is merged.

till-m (approved):

till-m:
@lm314 sorry for the delay. Could you please pull upstream changes?

PR #445, opened 2023-10-10
Update various docstrings, add workflow to check docstrings

till-m:
see #427

### DIFF
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -226,7 +226,7 @@ class BayesianOptimization(Observable):
         self.dispatch(Events.OPTIMIZATION_STEP)
 
     def probe(self, params, lazy=True):
-        """Evaluate the function on the given points.
+        """Evaluate the function at the given points.
 
         Useful to guide the optimizer.
 
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -1,3 +1,7 @@
+"""Main module.
+
+Holds the `BayesianOptimization` class, which handles the maximization of a
+function over a specific target space."""
 import warnings
 
 from bayes_opt.constraint import ConstraintModel
@@ -12,17 +16,25 @@ from sklearn.gaussian_process import GaussianProcessRegressor
 
 
 class Queue:
+    """Queue datastructure.
+
+    Append items in the end, remove items from the front.
+    """
+
     def __init__(self):
         self._queue = []
 
     @property
     def empty(self):
+        """Check whether the queue holds any items."""
         return len(self) == 0
 
     def __len__(self):
+        """Return number of items in the Queue."""
         return len(self._queue)
 
     def __next__(self):
+        """Remove and return first item in the Queue."""
         if self.empty:
             raise StopIteration("Queue is empty, no more objects to retrieve.")
         obj = self._queue[0]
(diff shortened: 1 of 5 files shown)


bayes_opt/bayesian_optimization.py
```diff
@@ -233,7 +277,13 @@ def suggest(self, utility_function):
...
 
     def _prime_queue(self, init_points):
-        """Make sure there's something in the queue at the very beginning."""
+        """Ensure the queue is not empty.
+
+        Parameters
+        ----------
+        init_points: int
+            Number of parameters to prime the queue with.
+        """
```
leandrobbraga, line 286: Is this *really* what this function does? The term "prime" gives me the impression that this is preparing/driving the queue to the start point.

leandrobbraga (approved):

bayes_opt/bayesian_optimization.py
```diff
@@ -167,30 +182,53 @@ def __init__(self,
...
+        constraint_value: float or None
+            Value of the constraint function at the observation, if any.
+        """
         self._space.register(params, target, constraint_value)
         self.dispatch(Events.OPTIMIZATION_STEP)
 
     def probe(self, params, lazy=True):
-        """
-        Evaluates the function on the given points. Useful to guide the optimizer.
+        """Evaluate the function on the given points.
```
leandrobbraga, line 229:
```suggestion
        """Evaluate the function at the given points.
```

leandrobbraga (changes requested):

bayes_opt/bayesian_optimization.py
till-m: The code is pretty short and fits here:
```python


    def _prime_queue(self, init_points):
        """Ensure the queue is not empty.

        Parameters
        ----------
        init_points: int
            Number of parameters to prime the queue with.
        """
        if self._queue.empty and self._space.empty:
            init_points = max(init_points, 1)

        for _ in range(init_points):
            self._queue.add(self._space.random_sample())
```

So what happens is a) if the queue is empty and there are no points registered to the target space it samples at least one point randomly from the space and adds it to the queue. I'm open to changing the description (or function name) to something that is more intuitive. Any suggestions?

> The term "prime" gives me the impression that this is preparing/driving the queue to the start point.

Not sure what exactly you mean by this, I think the `prime` is used here as a stand-in for `prepare`.

bwheelz36:
Hey @till-m, sorry I don't have time to review in detail - great work on pushing this forward!

Just a query on the pydocstring - I just had a look at running it locally and it seems extremely sensitive - producing hundreds and hundreds of errors. I'm not sure if they would all have to be fixed for the job to pass? 
A lot about things I don't really care about, like missing spaces, missing full stops etc. It might be possible to configure this so it only catches things we really care about - which for me is (1) docstring missing and (2) undescribed parameters.

till-m:
@bwheelz36, it's indeed super sensitive. I think there's two possibilities:
1) Once all the docstrings adhere to the numpy standard, there should ideally not be a lot of work in maintaining that.
2) use `--select=CODE01,CODE02,ETC` to select only specific things to check against.

I think there's an argument to be made in staying consistent with other scientific packages, lots of which use numpy's style. At the same time, I kind of agree that it seems to be overly pedantic. I would be fine with going either direction.

bayes_opt/bayesian_optimization.py
leandrobbraga: I think it's fine

Issue #439, opened 2023-07-22
Implement `gp_hedge` acquisition function


pfebrer:
**Is your feature request related to a problem? Please describe.**

I'm trying to migrate from `scikit-optimize` to this package because it seems more actively maintained. However, in `scikit-optimize` the default acquisition function was `gp_hedge`, a combination of the three acquisition functions implemented in this package, which is not implemented in this package.

**Describe the solution you'd like**
Would it be possible to add `gp_hedge` utility function to this package?

**References or alternative approaches**

The methodology is described in https://arxiv.org/pdf/1009.5419.pdf (Algorithm 2 in page 6).

And it is implemented in `scikit-optimize` loop here:
https://github.com/scikit-optimize/scikit-optimize/blob/a2369ddbc332d16d8ff173b12404b03fea472492/skopt/optimizer/optimizer.py#L538-L539

and here:

https://github.com/scikit-optimize/scikit-optimize/blob/a2369ddbc332d16d8ff173b12404b03fea472492/skopt/optimizer/optimizer.py#L596-L602

**Additional context**

I'm new to this field, so I might be missing some points like gp_hedge not being the optimal technique or being described initially in another paper.

**Are you able and willing to implement this feature yourself and open a pull request?**
- [ ] I'm not sure I would do it in the right way.

till-m:
Hi @pfebrer,

this would seem like a good addition to the project, in my opinion. However, compared to other acquisition functions, this one seems to be non-trivial to integrate into the existing codebase, primarily because the process is stateful through the `gains` vector. In all likelihood this will mean that the resulting optimization process...

- ...will not be serializable.
- ...will have no clear separation between suggest/evaluate/register.

I currently don't have the bandwidth to implement this either. You could give it a try and see if you can work around these issues in a clever way.

PS: maybe I misread the paper -- I really just scanned the pseudocode on p.6 -- and this is not actually an issue.

pfebrer:
Indeed that's the scary part about it, the acquisition function needs to keep a state.

But I don't know if it's really incompatible with the current code structure. You already define aquisition functions by initializing a python object, e.g.:

```python
utility = UtilityFunction()
```

So that object could keep a state that it's updated when you pass the utility function to the `suggest` method.

till-m:
[...]

PR #447, opened 2023-10-22
Acquisition function API redesign

till-m:
Redesign the way acquisition functions work.

Content:
- Deletes the `UtilityFunction` class and replaces it with a generic `AcquisitionFunction` base class
- Delete `acq_max`, let each acquisition function handle the maximization itself
- Adds `UpperConfidenceBound`, `ProbabilityOfImprovement` and `ExpectedImprovement` as possible utility functions
- Adds wrapper acquisition function `GPHedge` to find which of (a combination of) UCB/PoI/EI are best suited for a problem #439 
- Adds wrapper acquisition function `KrigingBeliever` to allow for async optimization #347 #365 
- Disallow constrained optimization for UCB -- the constrained optimization relies on the acquisition function being lower-bounded by zero, otherwise the rescaling doesn't make so much sense. While the old implementation seems to have worked sufficiently well, this is more principled.
- Set EI as standard acquisition function for constrained improvement
- Make the acquisition function a property of the optimizer (this should mean the the optimizer as a whole is serializable, including the acquisition function, though I haven't actually tested this). This change is **breaking** as it means `suggest`, `maximize` etc. don't take an acquisition function argument anymore.

Why?
The old `UtilityFunction` and `acq_max` were the most inflexible part of this package and their design was somewhat convoluted (e.g. providing a `kappa` parameter when using UCB). Now it should be much easier to add custom acquisition functions.

I will add more documentation later in the process, I think this shouldn't be merged before the docstring overhaul is done anyways.


till-m:
Some more changes:
- Add `ConstantLiar`
- fix a bug in `exploration_decay`
- fix a bug in `KrigingBeliever`, where the GP was fitted on the wrong data
- Adapted the `async_optimization.py` script to showcase `ConstantLiar`
- Reran some notebooks

@fmfn @bwheelz36 @leandrobbraga I will wait with merging this until the docstrings are finished, but if you'd like to give some feedback already, please do.
@pfebrer this contains GPHedge, if you'd like to give some feedback on it.
@rizhiy I added the constant liar and the kriging believer for async optimization, as well as a script based on your code in `examples/async_optimization_dummies.py`. Ideally this will eventually become a notebook. According to [this](https://www.cs.ubc.ca/labs/algorithms/EARG/stack/2010_CI_Ginsbourger-ParallelKriging.pdf), lying is the better strategy, but I wasn't sure whether it would make sense for constrained optimization, so I added the believer for that case.

bwheelz36:
Wow this is a huge amount of work. Impressive! Before we start review can you resolve the merge conflicts and make sure all the tests pass?

till-m:
> Before we start review can you resolve the merge conflicts and make sure all the tests pass?

Will do 👍  ideally I want to wait until the docs are finished to only deal with one conflict. After that I will mark the PR as ready to review :)

bwheelz36:
@till-m - will be a breaking redesign? Not necessarily against it, just want to be clear.

till-m:
> @till-m - will be a breaking redesign? Not necessarily against it, just want to be clear.

There will be some necessary breaking changes, primarily to workflows that change the acquisition function -- this is unavoidable, since the old `UtilityFunction` object has been replaced. I also (intentionally) modified the way the acquisition function is changed -- it is now a property of the optimizer (i.e. not an argument to maximize/suggest). I understand that this can be annoying for some users, but I think in the long run it will be exactly those users that will profit from this change.

If someone just calls `.maximize` nothing should change. Similarly, `suggest`-`probe`-`register` w/o messing with the acquisition function will not change either.

till-m:
@bwheelz36 I know this is a lot. I'm sorry about that 😬

bayes_opt/bayesian_optimization.py
```diff
@@ -347,29 +339,20 @@ def maximize(self,
         self.dispatch(Events.OPTIMIZATION_START)
         self._prime_queue(init_points)
 
-        old_params_used = any([param is not None for param in [acq, kappa, kappa_decay, kappa_decay_delay, xi]])
+        old_params_used = any([param is not None for param in [kappa, kappa_decay, kappa_decay_delay, xi]])
```
till-m, line 342: Maybe this would be a good opportunity to drop these parameters from the call signature.

bayes_opt/acquisition.py
```diff
@@ -0,0 +1,854 @@
...
+    pass
+
+
+class NoValidPointRegisteredError(Exception):
+    """Raised when an acquisition function depends on previous points but none are registered."""
+
+    pass
+
+
+class TargetSpaceEmptyError(Exception):
```
till-m, line 27: I'm not entirely sure whether it is appropriate to define such custom errors, or whether it would be better to use `ValueError`s and `TypeError`s here.

bayes_opt/acquisition.py
```diff
@@ -0,0 +1,854 @@
...
+        self._fit_gp(gp=gp, target_space=dummy_target_space)
+        x_max = self.base_acquisition.suggest(gp, dummy_target_space, n_random=n_random, n_l_bfgs_b=n_l_bfgs_b, fit_gp=False)
+
+        # Register the suggested point as a dummy
+        self.dummies.append(x_max)
+
+        return x_max
+
+
+class GPHedge(AcquisitionFunction):
```
till-m, line 746: I didn't want to go overboard and add too many acquisition functions. `GPHedge` and the `ConstantLiar` seemed appropriate. EIpu (#393) is additionally showcased in the notebook.

bayes_opt/bayesian_optimization.py
```diff
@@ -137,6 +136,14 @@ def __init__(self,
         self._allow_duplicate_points = allow_duplicate_points
         self._queue = Queue()
 
+        if acquisition_function is None:
```
bwheelz36, line 139: Do we want to use a default here, or do we want to throw an error - that is, if people adapt the new version, we force them to update their codes? I feel like we got ourselves in trouble in the past by silently using defaults?

bayes_opt/acquisition.py
bwheelz36: I'm not sure either actually. I've been wondering this recently myself.

On the plus side, custom errors tend to be clearer. However, for an external program/user to catch these exceptions can be annoying. 

After a bit of reading, I think that if we expect that an external program may want to handle the errors then using standard errors may be better - but if 
1. the error would be expected to be handled internally
2. OR not handled at all (that is: crash is the desired result), 
3. OR there is no standard error that can reasonably be applied
THEN a custom error may be better?

- https://softwareengineering.stackexchange.com/questions/381893/is-it-a-good-practice-to-use-self-defined-exception
- https://stackoverflow.com/questions/45735280/how-to-catch-custom-exception-in-python

bayes_opt/bayesian_optimization.py
till-m: The idea was to mirror [this line](https://github.com/bayesian-optimization/BayesianOptimization/blob/f2c720e76c6de2d5b334a3e971acdbc28ab37ed9/bayes_opt/bayesian_optimization.py#L304) in `.maximize`. So for someone not supplying any acquisition function, the behaviour should remain the same. On the other hand, if someone is used to supply an acquisition function to `.maximize` or `.suggest`, it would now crash with a `TypeError: got an unexpected keyword argument`.

till-m: there's also the option of adding an `acquisition_function` argument to `maximize` and explicitly raise an error if someone supplies it.

bayes_opt/acquisition.py
```diff
@@ -0,0 +1,854 @@
...
+    pass
+
+
+class TargetSpaceEmptyError(Exception):
+    """Raised when the target space is empty."""
+
+    pass
+
+
+class AcquisitionFunction():
```
bwheelz36, line 33: could use the [ABC](https://docs.python.org/3/library/abc.html) functionality to define the abstract base class, as well as any abstract methods using `@abstractmethod`

bayes_opt/acquisition.py
```diff
@@ -0,0 +1,854 @@
...
+                self.random_state = random_state
+            else:
+                self.random_state = RandomState(random_state)
+        else:
+            self.random_state = RandomState()
+        self.i = 0
+
+    def base_acq(self, *args, **kwargs):
+        """Provide access to the base acquisition function."""
+        raise NotImplementedError
```
bwheelz36, line 54: I think that `@abstractmethod` would be more appropriate here. `NotImlpementedError` to my understanding is intended to be used in situations where the functionality WILL eventually be implemented.

bayes_opt/bayesian_optimization.py
bwheelz36: I think the current behavior is probably good, I like that it throws an error in more complex cases.

tests/test_acquisition.py
```diff
@@ -0,0 +1,279 @@
+import pytest
+import numpy as np
+from bayes_opt import acquisition
+from sklearn.gaussian_process import GaussianProcessRegressor
+from bayes_opt.target_space import TargetSpace
+from bayes_opt.constraint import ConstraintModel
+from scipy.spatial.distance import pdist
+from bayes_opt import acquisition 
+# TODO: Add tests that checks that the acq_max actually returns the maximum
```
bwheelz36, line 9: should we open a new issue for this rather than leave the ToDo in source?

bwheelz36 (approved):
- great work @till-m. I have very few comments. I agree that this seems cleaner and more maintainable. I haven't used many different acquisition functions, so I don't have any comments on the specific implementations. 
- nitpicking, there's a number of lines `if self.y_max is None:` which are not covered by a test case. but coverage is >95% which is already extremely high, so if you don't think these need to be covered I'm fine with it.
- It is a breaking change, so when we release the next version it should be `2.0.0`
- Given the significance of the change, ideally we'd get a review from multiple people before merging. But if no one else chimes in after a week or so I think you should merge.

bayes_opt/bayesian_optimization.py
bwheelz36: do it! :-)

bayes_opt/acquisition.py
till-m: Thanks for the links! Some good answers in there.
My personal conclusion is this: I think the goal here is to crash the program with an informative error, so that the user knows what to fix. Every single one of these Errors have a clear fix, so I think it's reasonable to have custom classes here and not to conflate them with something potentially more abstract. I would lean to keeping them, though I don't have strong feelings.

bayes_opt/bayesian_optimization.py
till-m: done.

bayes_opt/acquisition.py
till-m: good call, I changed it as you suggested

bayes_opt/acquisition.py
till-m: done.

tests/test_acquisition.py
till-m: I went ahead and added these tests + deleted the line. They're basic tests (optimizing a 2d quadratic function), but I guess that's all that's needed.

till-m:
Coverage is fixed now! :)

Agree on all other points. Thanks for the review! :)

PR #457, opened 2024-01-24
Docstring overhaul

### LINKED ISSUE #427: settle on a docstring format and update all docstrings

till-m:
Resolves #427 #434 
- Rewrites most docstrings
- Uses `numpydoc`
- Include README.md in the docs
- Change links in README.md to link preferrably to docs webpage
- Change github links in README to link to the organization repo
- Replaces colors with colorama #434 
- Migrate images from `examples/` to `static/`


till-m:
It looks like something was messed up in all the merges. I will close this PR for now and have a look when I'm back on vacation.

till-m:
looking back now, I think making a big PR on a seperate branch was not a great idea...

Some of the docstring rules imposed by the `numpydocs` convention are a bit silly imo, but I guess there is an argument to be made for consistency.
Either way, this should hopefully give us a solid base of docstrings to work with. I think we can then refine how the documentation should look in the future.

perezed00 (approved):
So far everything I've seen looks good. I took the closest look at domain reduction and target space changes.

bayes_opt/domain_reduction.py
```diff
@@ -180,20 +212,47 @@ def _trim(self, new_bounds: np.array, global_bounds: np.array) -> np.array:
...
+        """Transform the bounds of the target space.
+        
+        Parameters
+        ----------
+        target_space : TargetSpace
+            TargetSpace this DomainTransformer operates on.
+        
+        Returns
+        -------
+        dict
```
perezed00, line 254:
```suggestion
        bounds : dict
```

bayes_opt/target_space.py
```diff
@@ -209,11 +297,17 @@ def register(self, params, target, constraint_value=None):
...
             else:
-                raise NotUniqueError(f'Data point {x} is not unique. You can set "allow_duplicate_points=True" to '
-                                     f'avoid this error')
+                raise NotUniqueError(f'Data point {x} is not unique. You can set'
+                                     ' "allow_duplicate_points=True" to avoid this error')
+
+        # if x is not within the bounds of the parameter space, warn the user
+        if self._bounds is not None:
+            if not np.all((self._bounds[:, 0] <= x) & (x <= self._bounds[:, 1])):
+                warn(f'\nData point {x} is outside the bounds of the parameter space. ', stacklevel=2)
```
perezed00, line 310: lines `307-310` appear to be duplicates of `312-316`

bayes_opt/target_space.py
```diff
@@ -237,17 +331,15 @@ def register(self, params, target, constraint_value=None):
...
     def probe(self, params):
-        """
-        Evaluates a single point x, to obtain the value y and then registers the
-        point and its evaluation.
+        """Evaluate the target function on a point and register the result.
 
         Notes
         -----
-        If x has been previously seen returns a cached value of y.
+        If `params` has been previously seen returns a cached value of `y`.
```
perezed00, line 338: I'm not sure I see this functionality in this function. I think its calculating and returning target every time.

bayes_opt/domain_reduction.py
perezed00: In general, should we have the fields under "Returns" have the following format? object_name : object_type ?

I understand that just reporting the type suffices, but I think a name could be more clear. Thoughts?

perezed00:
@till-m please re-request my review. My previous approval of all commits was in error.

bayes_opt/target_space.py
till-m: My bad, this must have happened when trying to fix a merge conflict.

bayes_opt/domain_reduction.py
till-m: I'm fine with either -- imo it _should_ be clear from the function name what it returns ideally, but it doesn't add much overhead.

bayes_opt/target_space.py
till-m: You're right, this was removed in #372. Should we adjust the behaviour here to return the cached value or change the docstring to reflect the current code (where this would raise an error).

till-m: In either case, we should probably check whether a point is already registered before evaluating the target function potentially unnecessarily?

bayes_opt/domain_reduction.py
bwheelz36: my vote: it's only necessary to specify return type, not return name since the name of the returned object depends on how the function is called.

.github/workflows/build_docs.yml
```diff
@@ -29,6 +29,8 @@ jobs:
           pip install nbsphinx
           pip install sphinx_rtd_theme
           pip install jupyter
+          pip install numpydoc
```
bwheelz36, line 32: I guess numpydoc isn't required in this CI file anymore?

.github/workflows/check_docstrings.yml
```diff
@@ -0,0 +1,30 @@
+# This workflow will install Python dependencies and check docstrings with a single version of Python
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Check docstrings
```
bwheelz36, line 4: we could consider integrating this into the tests job, but I think for now leave as is :-)

bwheelz36:
Nice work @till-m 
I won't review all docstrings, but I see you implemented an automatic checker which I think is absolutely the right approach.

- I'm not familiar with `numpy-docs` - does it allow for configuration to ignore trivial errors?
- Will it detect missing docstrings/parameter descriptions in public methods? this is the main thing I'd really like to make sure we nail down. 
- What I have done on other code is use [flake8-docstrings](https://github.com/pycqa/flake8-docstrings) and then set it to ignore the largely spurious errors it throws up (things like missing spaces, full stops etc.)

till-m:
Hey @bwheelz36,

> I'm not familiar with numpy-docs

I think we don't actually need this to check for convention. I will remove it.

> Will it detect missing docstrings/parameter descriptions in public methods? this is the main thing I'd really like to make sure we nail down.

It will detect missing docstrings as-is
```
bayes_opt/bayesian_optimization.py:369 in public method `set_bounds`:
        D102: Missing docstring in public method
```

I didn't realize, but missing parameters needs to be explicitly enabled. I will enable it. Note that this does not detect missing docstrings of class documentation, or missing parameters in class documentation. This is due to the way `__init__` is handled.

> does it allow for configuration to ignore trivial errors?

You can additionally ignore errors by passing ` --add-ignore <error code>` in the call. Unfortunately one cannot ignore individual lines (similar to `noqa`). The tool you sent is actually a wrapper around `pydocstyle`, so it behaves the same (except it additionally produces a number of linting issues). If we want to additionally add linting rules, we could also do that, though in that case I would suggest to use `ruff` which has the advantage of being both very comprehensive and very fast. However, if we want to use `ruff` we should merge #458 into this branch, as it requires the `pyproject.toml`.

.github/workflows/build_docs.yml
till-m: removed!

bayes_opt/target_space.py
till-m: I changed the behaviour to be in line with the docstring again.

bayes_opt/target_space.py
```diff
@@ -136,23 +205,44 @@ def mask(self):
...
+
+        Parameters
+        ----------
+        params : dict
+            a single point, with len(x) == self.dim.
+
+        Returns
+        -------
+        np.ndarray
+            Representation of the parameters as dictionary.
```
perezed00, line 220:
```suggestion
            Representation of the parameters as an array.
```
I think the previous version was a typo

perezed00:
I took a second look at revisions in domain reduction, target space files, and several other `.py` files. However, I haven't looked at `util.py`, `logger.py`, or `constraint.py`

Within the files I checked, the suggestions I made were all rather small and generally things look good. Great work on this.

bayes_opt/target_space.py
till-m: Good catch, thanks!

bwheelz36:
@till-m - ruff looks really cool. I'm keen to try it out in some projects - but agree right now is not the time or place for this one :-)

till-m:
From my side, this PR is mostly good to go, except for a review of the README changes. Does anyone else have any other hold-ups?

README.md
```diff
@@ -277,54 +271,71 @@ load_logs(new_optimizer, logs=["./logs.log"]);
...
+    author={Gardner, Jacob R and Kusner, Matt J and Xu, Zhixiang Eddie and Weinberger, Kilian Q and Cunningham, John P},
+    booktitle={ICML},
+    volume={2014},
+    pages={937--945},
+    year={2014}
 }
 ```
 
-# Dependencies
+### Dependencies
```
bwheelz36, line 333: suggest we remove this - it should be clear from the packaging tools. it is even now, and will be more so when we switch to poetry

README.md
```diff
@@ -277,54 +271,71 @@ load_logs(new_optimizer, logs=["./logs.log"]);
...
-
-      conda install -c conda-forge bayesian-optimization
-
-The bleeding edge version can be installed with:
-
-    pip install git+https://github.com/fmfn/BayesianOptimization.git
+## Installing from GitHub
+To install directly from master, simply run
+```
+pip install https://github.com/bayesian-optimization/BayesianOptimization/archive/master.zip
```
bwheelz36, line 279: we added this while we couldn't access pypi, which is fixed now. we could probably remove now?

README.md
```diff
@@ -1,16 +1,18 @@
 <div align="center">
-  <img src="https://github.com/fmfn/BayesianOptimization/blob/master/examples/func.png"><br><br>
+  <img src="https://raw.githubusercontent.com/bayesian-optimization/BayesianOptimization/master/static/func.png"><br><br>
```
bwheelz36, line 2: if I look at your branch, it appears this image is missing now? can you double check this link still works?

bwheelz36:
> From my side, this PR is mostly good to go, except for a review of the README changes. Does anyone else have any other hold-ups?

Just took a look at readme - made a few minor suggestions, but looks great to me!

README.md
till-m: done.

README.md
till-m: done.

till-m:
If someone would still like to have a look, please let me know. Otherwise I will merge this early next week.

till-m:
Thanks everyone for the help! Glad to see this is done -- even if it isn't perfect right now, we can work to make it better in the future! :)

@bwheelz36 looks like the image renders correctly: [link](http://bayesian-optimization.github.io/BayesianOptimization/quickstart.html)

PR #479, opened 2024-06-20
Fix 478, fix data out-of-sync issue in target space

### LINKED ISSUE #478: Logging for multiple constraints to be fixed
**Describe the bug**
Using 'scipy == 1.11.1' with 'bayesian-optimization == 1.4.0' I see that I cannot log if multiple constraints are defined:

```python
def constraint_function_2_dim(x, y):
    return np.array([
        - np.cos(x) * np.cos(y) + np.sin(x) * np.sin(y),
        - np.cos(x) * np.cos(-y) + np.sin(x) * np.sin(-y)]).tolist()

constraint_lower = np.array([-np.inf, -np.inf]).tolist()
constraint_upper = np.array([0., 0.]).tolist()

constraint = NonlinearConstraint(constraint_function_2_dim, constraint_lower, constraint_upper)
    f=target_function,
    constraint=constraint,
    pbounds=pbounds,
    verbose=0, # verbose = 1 prints only when a maximum is observed, verbose = 0 is silent
    random_state=1,
)
 
 
logger = JSONLogger(path="./log_multiple_constraints")  # writes logs.json
optimizer.subscribe(Events.OPTIMIZATION_STEP, logger)
 
optimizer.maximize(
    init_points=2,
    n_iter=10,
)
```

getting the error:
`Object of type ndarray is not JSON serializable`

We suggest to fix this way:
modify `target_space.py` this way:

```
def res(self):

        if self._constraint is None:
            params = [dict(zip(self.keys, p)) for p in self.params]

            return [
                {"target": target, "params": param}
                for target, param in zip(self.target, params)
            ]
        else:
            params = [dict(zip(self.keys, p)) for p in self.params]

            return [
                {
                    "target": target,
                    "constraint": constraint_value.tolist(),
                    "params": param,
                    "allowed": allowed
                }
                for target, constraint_value, param, allowed in zip(
                    self.target,
                    self._constraint_values,
                    params,
                    self._constraint.allowed(self._constraint_values)
                )
            ]
```

(basically, adding "tolist()" to constraint_value in the returned list)

Thanks a lot!

till-m:
Fixes #478 + adds a test case for this.

Additionally, create copies during `.register()`. This makes sure that if something goes wrong during the registration phase, the target space remains unchanged. Previously, this could lead to a differing number of entries in `targets`/`params`/`constraints`, depending on where the error occurred.


till-m:
The `fontsize` problem seems to be related to a change in matplotlib, not to the other changes here.

bwheelz36:
will review today :-)

bwheelz36 (approved):
lgtm. no comments :-)

PR #485, opened 2024-06-25
Move static folder

### LINKED ISSUE #467: Move `static` inside `docsrc`
Very minor detail, but to avoid clutter, we could move the static images inside docsrc. Would require updating all the relative image paths - I'm happy to do it if you agree @till-m?

bwheelz36:
this turned out to be a lot more work/ changes than I expected...

- Fixes #467 
- Updates the build_docs CI job to use pyproject.toml and poetry
- also added a link to the scikit learn docs on different GP kernels to advanced tour notebook


.github/workflows/build_docs.yml
```diff
@@ -22,21 +22,15 @@ jobs:
...
-          pip install myst-parser
-      - name: Install package
-        run: |
-          pip install -e .
+          poetry install --with dev,nbtools
       - name: build sphinx docs
         run: |
           cd docsrc
-          make github
+          poetry run make github
```
till-m, line 33: Is there a reason why we need to run these commands with poetry?

docsrc/Makefile
```diff
@@ -15,10 +15,10 @@ BUILDDIR      = ../docs
 .PHONY: help Makefile
 
 github:
-	@cp ../README.md .
-	@cp -r ../static .
+# 	@cp ../README.md .
 	@make html
 	@cp -a ../docs/html/. ../docs
+	@ cp -r ../docsrc/ ../docs
```
till-m, line 21: Nitpick

```suggestion
	@cp -r ../docsrc/ ../docs
```

till-m (approved):

.github/workflows/build_docs.yml
bwheelz36: No, just the easiest way to activate the venv. You don't like it?

till-m: Ah I see otherwise the job fails. All good, thanks

Issue #488, opened 2024-07-04
Potential for ANSI colours to remain active


d7919:
In `_step` of logger, the colour remains active after the line is displayed, i.e. it is not reset.

https://github.com/bayesian-optimization/BayesianOptimization/blob/3a5e56c7d0f96affdebfebc318b02f71f0462be3/bayes_opt/logger.py#L166

If the user's function has any output or if the system displays any output this will inherit the colour of the last output. This can also potentially cause issues if the user kills the job mid-run leaving the colour code active.

Changing this line to 

~~~~python
return "| " + " | ".join([colour + cells[i] + Fore.RESET for i in range(len(cells))]) + " |" 
~~~~

is perhaps a bit "safer"  and has the minor advantage of not colouring the `|` column separators.

On a side note, it would be helpful if the user could:

1. Turn off the colouring (if capturing output the escape codes can enter the log file).
2. Choose the two colours in use -- for me the default Black is unreadable due to the colour of my terminal background.

till-m:
Hi @d7919,

regarding the cause for the issue, good catch, this should indeed be fixed. Thanks for providing a PR!

> side note

you raise some good points:
> Choose the two colours in use -- for me the default Black is unreadable due to the colour of my terminal background.

I think it would make sense to replace `Fore.BLACK` with `Fore.RESET` in general. What do you think? This would probably alleviate the issue for most cases and be relatively low effort. However, if you feel strongly about making the `ScreenLogger` more flexible/customizable (turning colours off or using different colours), and are willing submit a PR for this, I will gladly review and merge it :)

till-m:
I will close this for now

PR #489, opened 2024-07-04
Update logger.py to reset color escape sequence in between values

d7919:
Addresses issue #488 by resetting colour code after each value in `logger::_step`.

Also resets the use of red in target_space when warning about duplicate points being added.

### DIFF
--- a/bayes_opt/logger.py
+++ b/bayes_opt/logger.py
@@ -163,7 +163,7 @@ class ScreenLogger(_Tracker):
         for key in instance.space.keys:
             cells.append(self._format_number(res["params"][key]))
 
-        return "| " + " | ".join([colour + cells[i] for i in range(len(cells))]) + " |"
+        return "| " + " | ".join([colour + cells[i] + Fore.RESET for i in range(len(cells))]) + " |"
 
     def _header(self, instance):
         """Print the header of the log.
--- a/bayes_opt/logger.py
+++ b/bayes_opt/logger.py
@@ -46,7 +46,10 @@ class ScreenLogger(_Tracker):
 
     _default_cell_size = 9
     _default_precision = 4
-
+    _colour_new_max = Fore.MAGENTA
+    _colour_regular_message = Fore.RESET
+    _colour_reset = Fore.RESET
+    
     def __init__(self, verbose=2, is_constrained=False):
         self._verbose = verbose
         self._is_constrained = is_constrained
@@ -136,7 +139,7 @@ class ScreenLogger(_Tracker):
             return s[:self._default_cell_size - 3] + "..."
         return s
 
-    def _step(self, instance, colour=Fore.BLACK):
+    def _step(self, instance, colour=_colour_regular_message):
         """Log a step.
 
         Parameters
@@ -145,7 +148,7 @@ class ScreenLogger(_Tracker):
             The instance associated with the event.
             
         colour :
-            (Default value = Fore.BLACK)
+            (Default value = _colour_regular_message, equivalent to Fore.RESET)
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


till-m (approved):
Hi @d7919,

thanks for raising the issue and providing a solution. I had a quick look and the fix looks good.

As mentioned in a comment on your issue, I think it would also make sense to replace `Fore.BLACK` with `Fore.RESET` in general. If you agree that this is a good solution we could do that as part of this PR. Otherwise I will merge this PR within a week, provided there are no reservations from anyone else.

d7919:
Hi @till-m,

Thanks for the review. Replacing `Fore.BLACK` with `Fore.RESET` is a good idea as a "safe" default. 

I do think having a way to disable entirely could be useful for saving the screen logger output to file as well as to allow the user to avoid issues should they be running in a terminal with a magenta background (if there's anyone that does this!). 

Currently there's no user exposure of the logger (created in `_prime_subscriptions`) . If this creation was moved to near the end of `BayesianOptimization::__init__` and stored as `self.logger`  it should provide some exposure. Then by providing a method on `ScreenLogger` of the form:

~~~~python
def disable_colors(self):
     self.NEW_MAX_COLOR = ''
     self.DEFAULT_COLOR=''
     self.RESET_COLOR=''
~~~~

and changing the uses of Fore to `NEW_MAX_COLOR` etc. (which would be initialised in the `__init__` to the current values from Fore) then there's a simple method to strip out all escape codes and a way for users to customise these if they really want. 

This does feel like it might be exposing / adding a bit much -- I'm not really a python developer so there may be neater ways to do this, but I thought exposing the logger might be nicer than doing it all through init arguments which get passed down the call chain.

till-m:
Hi @d7919,

I'm not really an expert regarding the Observer pattern that's in use here, so I can't really see a nice & clean way of handling things. It can be handled however:

```python
from bayes_opt.logger import ScreenLogger
from bayes_opt.event import Events

def black_box_function(x, y):
    return -x ** 2 - (y - 1) ** 2 + 1

pbounds = {'x': (2, 4), 'y': (-3, 3)}

    f=black_box_function,
    pbounds=pbounds,
    verbose=2, # verbose = 1 prints only when a maximum is observed, verbose = 0 is silent
    random_state=1,
)

logger = ScreenLogger()
logger._default_cell_size = 15 # changing something to show that it works

# the following needs to happen before any call to `.maximize`
# otherwise there will be two loggers associated with the optimizer.
for e in [Events.OPTIMIZATION_START, Events.OPTIMIZATION_STEP, Events.OPTIMIZATION_END]:
    optimizer.subscribe(e, logger)

optimizer.maximize(
    init_points=2,
    n_iter=3,
)
```

> This does feel like it might be exposing / adding a bit much

IMO as long as things are handled by the `ScreenLogger` and abstracted from the `BayesianOptimization` class it would be fine.

Let me know how you wish to handle things. I'm fine with merging this PR as-is, but I would also be open to adding the options discussed above.

d7919:
Hi @till-m,

Maybe best to leave this PR as-is and leave any further modification of colourising output to the future. 

I think the appearance of colours outside of the logger would suggest that to handle customisation of this most consistently probably requires a slight redesign (perhaps a simple container class through which colour escape codes can be obtained). There's also the fact that `BayesianOptimization` takes and stores `verbose`  but only uses this to pass through to `_get_default_logger` which may suggest a slight refactor might help separate concerns more completely? 

Ultimately this isn't something that really impacts on the utility of the code so no need to rush further changes in.

till-m:
I agree, I think logging might be due for an overhaul at some point, though it is not a priority.

I would still argue that replacing `Fore.BLACK` with `Fore.RESET` would be a good change for now. Are you interested in handling that with this PR? Otherwise I can make a seperate PR.

till-m:
Looks good to me!

till-m (approved):
LGTM!

till-m:
thanks for your contribution @d7919!

Issue #496, opened 2024-07-11
apply ruff


phi-friday:
**Is your feature request related to a problem? Please describe.**
For now, the quality of the code in this library is up to the contributors.
This has resulted in each piece of code having an inconsistent style, 
sometimes resulting in performance-impacting patterns.

Several libraries have been developed to fix this problem,
and I've seen some of them being used in CI.(`pydocstyle`)

**Describe the solution you'd like**
There are so many libraries to fix this problem that it was difficult to decide on a specific one.
However, since the release of `ruff`, many major libraries have decided to use it.
(`pandas`, `matplotlib`, `scipy`, `sklearn`, ...)

This is because it implies many rules that are supported by other libraries,
(`black`, `isort`, `pydocstyle`, `flake`, ...)
and it works the fastest.

It's also not hard to use.
Just write your code as before, and run `ruff format bayes_opt && ruff check bayes_opt`.
If you use `pre-commit`, it's even simpler.

It can be a little difficult to set up the `ruff` rules you want to use, while maintaining existing code style.
However, if you spend a little time on the initial setup, it's simple from there.

**References or alternative approaches**
If this feature was described in literature, please add references here. Additionally, feel free to add descriptions of any alternative solutions or features you've considered.

**Additional context**
https://github.com/astral-sh/ruff
[playground](https://play.ruff.rs/)

**Are you able and willing to implement this feature yourself and open a pull request?**
- [X] Yes, I can provide this feature.

PR #497, opened 2024-07-11
Feat apply ruff

phi-friday:
apply ruff example
related: #496


bayes_opt/acquisition.py
```diff
@@ -1,35 +1,35 @@
...
 from sklearn.gaussian_process import GaussianProcessRegressor
-from typing import Callable, List, Union, Tuple
-from copy import deepcopy
-from numbers import Number
+
+from .constraint import ConstraintModel
+from .target_space import TargetSpace
+
 
 class ConstraintNotSupportedError(Exception):
```
till-m, line 22: FYI I'll be moving these to `util.py` in #498.

bayes_opt/__init__.py
```diff
@@ -1,13 +1,16 @@
 """Pure Python implementation of bayesian global optimization with gaussian processes."""
-from .bayesian_optimization import BayesianOptimization, Events
-from .domain_reduction import SequentialDomainReductionTransformer 
-from .logger import ScreenLogger, JSONLogger
-from .constraint import ConstraintModel
-from . import acquisition
+
+from __future__ import annotations
```
till-m, line 3: I've had a look and it seems that this import is only needed when annotating with some not-yet-defined type. Is this necessary here?

bayes_opt/acquisition.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, random_state=None):
     @abc.abstractmethod
     def base_acq(self, *args, **kwargs):
         """Provide access to the base acquisition function."""
-        pass # pragma: no cover
+        # pragma: no cover
```
till-m, line 56: is excluding this line from coverage still necessary considering there's no actual statement made here?

bayes_opt/acquisition.py
```diff
@@ -263,15 +274,14 @@ def _l_bfgs_b_minimize(self, acq: Callable, bounds: np.ndarray, n_x_seeds:int=10
...
             min_acq = np.inf
-            x_min = np.array([np.nan]*bounds.shape[0])
+            x_min = np.array([np.nan] * bounds.shape[0])
 
         # Clip output to make sure it lies within the bounds. Due to floating
         # point technicalities this is not always the case.
         return np.clip(x_min, bounds[:, 0], bounds[:, 1]), min_acq
 
-    def _update_params(self) -> None:
+    def _update_params(self) -> None:  # noqa: B027
```
till-m, line 283: I'm not sure the way I tried to handle this case is correct. I want this function to
- a priori do nothing
- exist, so that I can call it during `.suggest`
- do something for specific functions (e.g. UBC)

bayes_opt/acquisition.py
```diff
@@ -299,17 +309,21 @@ class UpperConfidenceBound(AcquisitionFunction):
...
 
-    def __init__(self, kappa=2.576, exploration_decay=None, exploration_decay_delay=None, random_state=None) -> None:
-        super().__init__(random_state=random_state)
-        assert kappa >= 0, "kappa must be greater than or equal to 0."
+    def __init__(
+        self, kappa=2.576, exploration_decay=None, exploration_decay_delay=None, random_state=None
+    ) -> None:
+        if kappa < 0:
+            error_msg = "kappa must be greater than or equal to 0."
+            raise ValueError(error_msg)
```
till-m, line 317: It's good that you caught this case -- could you please also add a test case to cover it?

bayes_opt/constraint.py
```diff
@@ -48,13 +53,7 @@ def __init__(self, fun, lb, ub, random_state=None):
...
             raise ValueError(msg)
 
-        basis = lambda: GaussianProcessRegressor(
-            kernel=Matern(nu=2.5),
-            alpha=1e-6,
-            normalize_y=True,
-            n_restarts_optimizer=5,
-            random_state=random_state,
-        )
+        basis = partial(_create_gaussian_process_regressor, random_state=random_state)
```
till-m, line 56: Is this considered a better style compared to using `lambda`? It seems that this obscures to some degree what's happening.

bayes_opt/constraint.py
```diff
@@ -77,7 +76,7 @@ def eval(self, **kwargs: dict):
 
         Parameters
         ----------
-        \*\*kwargs :
+        **kwargs : any
```
till-m, line 79: I remember using the backslashes here because otherwise the docs would not render the double asterisks correctly -- something to keep in mind and to check.

bayes_opt/domain_reduction.py
```diff
@@ -45,24 +47,26 @@ class SequentialDomainReductionTransformer(DomainTransformer):
...
+        the same value is used for all parameters.
     """
 
     def __init__(
         self,
         gamma_osc: float = 0.7,
         gamma_pan: float = 1.0,
         eta: float = 0.9,
-        minimum_window: Optional[Union[List[float], float, Dict[str, float]]] = 0.0
+        minimum_window: list[float] | float | dict[str, float] | None = 0.0,
```
till-m, line 59: This could also be a `np.ndarray`, I think

bayes_opt/logger.py
```diff
@@ -155,18 +154,16 @@ def _step(self, instance, colour=_colour_regular_message):
         A stringified, formatted version of the most recent optimization step.
         """
         res = instance.res[-1]
-        cells = []
+        cells: list[str | None] = [None, None, None, None]
```
till-m, line 157: Maybe a tuple would be a more appropriate data structure here?

bayes_opt/observer.py
```diff
@@ -29,20 +35,19 @@ def _update_tracker(self, event, instance):
...
-            if (self._previous_max is None
-                    or current_max["target"] > self._previous_max):
+            if self._previous_max is None or current_max["target"] > self._previous_max:
                 self._previous_max = current_max["target"]
                 self._previous_max_params = current_max["params"]
 
     def _time_metrics(self):
         """Return time passed since last call."""
-        now = datetime.now()
+        now = datetime.now(_UTC)
```
till-m, line 50: I'm not sure I'm reading this correctly -- you're setting `now` to the UTC time? I think that's fine if we just use timedelta (i.e. never explicitly print the time) but otherwise we should probably use the timezone of the user's system.

bayes_opt/util.py
```diff
@@ -21,13 +25,13 @@ def load_logs(optimizer, logs):
...
     The optimizer with the state loaded.
-    
+
     """
     if isinstance(logs, str):
         logs = [logs]
 
     for log in logs:
-        with open(log, "r") as j:
+        with open(log) as j:  # noqa: PTH123
```
till-m, line 34: Would also be fine with using Pathlib here.

till-m (commented):
Hi @phi-friday,

thanks for this PR! I think it would really improve this codebase. I would merge #498 before merging this, but I think reviews here might still take some time.
I've commented on some things that I noticed. These are not necessarily requests to change things, it assume there's generally a solid reason for you handling things the way you do that I might not have seen.

Consider this sort of a preliminary review, I would like to have a look again once you feel the PR is ready.

Finally, if there are no objections from you or any other contributors, I might ask a coworker with some more SE experience to additionally review this PR. I don't think my background here is sufficient to provide solid feedback.

bayes_opt/__init__.py
phi-friday: Not necessary.
Automatically generated based on the `ruff` rule. 
You can change the `ruff` rule if you wish, or exclude it from the file.

bayes_opt/acquisition.py
phi-friday: After adding the rule to remove `pass`, I forgot to add a comment. I'll fix it.

bayes_opt/acquisition.py
phi-friday: Initially, I decorated it with `abstractmethod`.
But in the notebook file, I saw that there was an object that inherited from this class.

I switched to public methods and tried to define them in the notebook file, 
but public methods need document.

It was difficult to write a precise description of this method, 
so I decided to ignore the `B027` rule for now. 

In my opinion, it would be better to define it as an abstract method.

bayes_opt/acquisition.py
phi-friday: This is the code i added by removing `assert` from the existing code.
I'll add the test code soon.

bayes_opt/constraint.py
phi-friday: The use of `lambda` in `python` is not recommended except in some case. 
I have modified the code above to add the `ruff` rule accordingly.

If this code is obscure, there are two ways to go about it.
1. ignore the `ruff` rule in your existing code
2. change `lambda` to `def`

bayes_opt/constraint.py
phi-friday: I found that a backslashe was causing problems with `D417` and fixed it.
But now that I think about it, `pydocstyle` was fine.
It could be an issue with `ruff`.

I'll try to find another way.

bayes_opt/domain_reduction.py
phi-friday: Fixed some `type hints` that were being written while applying the `ruff` rule.
If you want to add `ndarray`, I'll add it.

bayes_opt/logger.py
phi-friday: I prefer `tuple`. But since `tuple` is immutable, I used `list`.

If I were to change to `tuple`, it would require a lot of changes in the underlying logic.

bayes_opt/observer.py
phi-friday: The `ruff` convention requires to specify `tzinfo`
when using the `.now()` method, so I applied `UTC`.

I decided on `UTC` because getting the exact local timezone is a bit tricky.
(although it would be simple to add other library).
I'll remove the rule from this part.

bayes_opt/util.py
phi-friday: I'll fix it soon.

bayes_opt/constraint.py
phi-friday: Changed `.eval()` to ignore `D417`.

bayes_opt/acquisition.py
phi-friday: I thought about it again.
it would be better to switch to public method(not abstract method) and add docs.
(The reason for switching to a public method is that 
the user needs to know that they can override the method.)

Users can override this method if they want.
In any case, it seems to be the best way to make sure that it runs normally.

till-m: Hi @phi-friday,

I also thought about it again.
I would suggest a different approach: Delete this function from the abstract base class, and retain it only for UBC/PoI/EI. Make it a public method. Call the function in the `.suggest` method for these 3, after the `super().suggest`. I think that's a clean solution -- what do you think?

till-m:
Could you ping me when you feel this is ready for a review?

bayes_opt/acquisition.py
phi-friday: If the user is responsible for using `_update_params` (will be renamed),
that seems best.

phi-friday:
> Could you ping me when you feel this is ready for a review?

@till-m 
Until the new `ruff` version is released, there's nothing else I'd like to modify.
I'm ready.

bayes_opt/acquisition.py
till-m: I've handled this in #498, [link](https://github.com/bayesian-optimization/BayesianOptimization/pull/498/commits/f31311601a6cee42c26219bc31b5ffe4331f3ceb).

phi-friday: > I've handled this in #498, [link](https://github.com/bayesian-optimization/BayesianOptimization/pull/498/commits/f31311601a6cee42c26219bc31b5ffe4331f3ceb).

method names are also clearer than before, which I think is a good thing. 
I'll rebase and push again when #498  is merged.

tests/test_notebooks_run.py
```diff
@@ -2,30 +2,34 @@
...
-        ep = ExecutePreprocessor(timeout=600, kernel_name='python3')
-        ep.preprocess(nb, {'metadata': {'path': Path(notebook_loc).parent}})
-    except Exception as e:
-        raise Exception(f'failed to run notebook {notebook_loc}')
-    print(f'success!')
+        ep = ExecutePreprocessor(timeout=600, kernel_name="python3")
+        ep.preprocess(nb, {"metadata": {"path": Path(notebook_loc).parent}})
+    except Exception as exc:
+        error_msg = f"failed to run notebook {notebook_loc}"
+        raise Exception(error_msg) from exc  # noqa: TRY002 # FIXME
```
till-m, line 24: Could be a `RuntimeError`, if you prefer. The point here is to just make sure the tests fail.

bayes_opt/acquisition.py
```diff
@@ -1,18 +1,25 @@
...
-from typing import Callable, List, Union, Tuple
-from copy import deepcopy
-from numbers import Number
-from .util import ConstraintNotSupportedError, NoValidPointRegisteredError, TargetSpaceEmptyError
+
+from bayes_opt.target_space import TargetSpace
+from bayes_opt.util import ConstraintNotSupportedError, NoValidPointRegisteredError, TargetSpaceEmptyError
+
+if TYPE_CHECKING:
+    from bayes_opt.constraint import ConstraintModel
```
till-m, line 22: Why is this condition necessary?

pyproject.toml
```diff
@@ -40,3 +41,10 @@ myst-parser = "^3.0.1"
 [build-system]
 requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
+
+[tool.coverage.report]
+exclude_lines = [
+    "pragma: no cover",
+    "if TYPE_CHECKING:",
+    "raise NotImplementedError",
```
till-m, line 49: If the `DomainTransformer` were an `abc.ABC` (which it probably should be), then we wouldn't need to exclude `NotImplementedError` anymore, correct? Or is this a standard line to include?

scripts/check.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
+set -ex
+
+poetry run ruff format --check bayes_opt tests
+poetry run ruff check bayes_opt tests
```
till-m, line 5: What exactly is the purpose of the scripts? How do you envision the workflow to be?

till-m (commented):
Looks good to me, generally. I added some comments to clarify things.

till-m:
@Panaetius Ralf, I would appreciate your feedback on this PR, if you have the time. No need to focus too much on the code (which should be functionally the same), the high-level things are what's important.

tests/test_notebooks_run.py
phi-friday: In my opinion, this `try... except` doesn't mean much.
It seems better to pass `notebook` as a parameter (using `pytest.mark.parametrize`).

bayes_opt/acquisition.py
phi-friday: This is because it is not used during runtime.
By importing it under the `TYPE_CHECKING` condition,
we can avoid circular reference errors that might occur in the future.

I haven't checked, but it is probably not necessary.

pyproject.toml
phi-friday: I forgot, I'll fix it soon.

scripts/check.sh
phi-friday: Although `pre-commit` is useful,
there are many times when I want to use commands directly before committing.
This is a simple script for those times.

If it works as intended,
This script should guarantee the same result as `pre-commit`.

pyproject.toml
phi-friday: I forgot.
https://github.com/bayesian-optimization/BayesianOptimization/blob/d305c1b6584432841a60120cf45e5a2f1b68bef0/bayes_opt/domain_reduction.py#L25
https://github.com/bayesian-optimization/BayesianOptimization/blob/d305c1b6584432841a60120cf45e5a2f1b68bef0/bayes_opt/domain_reduction.py#L29
These two places seem to be areas that are not tested and don't need to be tested.
But if `"raise NotImplementedError",` is removed, it will be reflected in the code coverage.

Should I leave them in, or comment `pragma: no cover`?

till-m: I think the move here would be to convert the `DomainTransformer` into an `abc.ABC` and use `@abstractmethod` here.  If you want, you can do that in this PR, otherwise we can leave things as-is.

phi-friday: > I think the move here would be to convert the `DomainTransformer` into an `abc.ABC` and use `@abstractmethod` here. If you want, you can do that in this PR, otherwise we can leave things as-is.

I'll fix it soon.

phi-friday: > > I think the move here would be to convert the `DomainTransformer` into an `abc.ABC` and use `@abstractmethod` here. If you want, you can do that in this PR, otherwise we can leave things as-is.
> 
> I'll fix it soon.

By the way, I have a slight problem with the code coverage (2 lines), is this okay?

스크린샷 2024-07-17 오후 11 47 18

tests/test_notebooks_run.py
phi-friday: > In my opinion, this `try... except` doesn't mean much. It seems better to pass `notebook` as a parameter (using `pytest.mark.parametrize`).

@till-m 
Should I fix it in this pr?

bayes_opt/logger.py
Panaetius: you could do 
```
num_cells = 4
cells: list[str|None] = [None] * num_cells
```
(or just `[None] * 4`)

scripts/check.sh
Panaetius: if it's just that, you can do `pre-commit run` to run it manually

bayes_opt/logger.py
phi-friday: > you could do
> 
> ```
> num_cells = 4
> cells: list[str|None] = [None] * num_cells
> ```
> 
> (or just `[None] * 4`)

Is `[None] * 4` more efficient than `[None, None, None, None]`?
If it is more efficient, then it is correct to modify it.

scripts/check.sh
phi-friday: > if it's just that, you can do `pre-commit run` to run it manually

That's right, that's what I usually do.
`poetry run pre-commit run --all-files --show-diff-on-failure`

But sometimes, when there are non-commit related modifications, 
`pre-commit` shows logs that are not needed.
And it seemed tricky to control.

pyproject.toml
till-m: I would handle this similar to the case of the acquisition function: Considering that the methods now have the @abstractmethod decorator you don't need to raise these Errors anyways. See [here](https://github.com/bayesian-optimization/BayesianOptimization/blob/bc9af23981cb22b98344424b5c2544b741b74938/bayes_opt/acquisition.py#L37) for comparison.

tests/test_notebooks_run.py
till-m: Yes, please!

pyproject.toml
phi-friday: > I would handle this similar to the case of the acquisition function: Considering that the methods now have the @AbstractMethod decorator you don't need to raise these Errors anyways. See [here](https://github.com/bayesian-optimization/BayesianOptimization/blob/bc9af23981cb22b98344424b5c2544b741b74938/bayes_opt/acquisition.py#L37) for comparison.

removing it from the code itself, I understand.



> If you don't need a base class with functionality, you can also use [protocols](https://docs.python.org/3/library/typing.html#typing.Protocol) in which case you don't need the NotImplementedError's

What do you think of the suggestion to use `Protocol` instead of `ABC`?

bayes_opt/logger.py
Panaetius: I don't think it's more efficient, if it's performance critical code, leave it as is.
Just a bit less repetition and more flexible if there ever could be different numbers of cells

pyproject.toml
till-m: I would prefer the ABC solution, because I'd rather not add another pattern to the code. I understand that the situation here is distinct from `acquisition.AcquisitionFunction` (because this one is literally nothing but a template) but I think with how small this codebase is they're not distinct enough to differentiate imo.

scripts/check.sh
till-m: Is there a reason why you did not similarly add a script to excute tests & calculate coverage, e.g. `poetry run pytest --cov-report xml --cov=bayes_opt/`?

till-m (approved):
I think the codecov bot is not working correctly and coverage is not being updated?

In any case, I think this PR looks good, insofar as I am able to judge it.

I would wait a bit more to see if anyone else wants to give a review, but then I would merge this soon.

scripts/check.sh
phi-friday: > Is there a reason why you did not similarly add a script to excute tests & calculate coverage, e.g. `poetry run pytest --cov-report xml --cov=bayes_opt/`?

I didn't add the script because the test took a long time.(>60s)

I wanted the scripts to be small enough that anyone could just run them whenever they wanted and see what happened.

till-m:
Thanks for your contribution!

PR #505, opened 2024-07-25
Fix apply enum

phi-friday:
When I looked at where `Events` is used, I realized that it would be better to inherit from `Enum`.

### DIFF
--- a/bayes_opt/event.py
+++ b/bayes_opt/event.py
@@ -2,12 +2,24 @@
 
 from __future__ import annotations
 
-from enum import Enum
+import sys
+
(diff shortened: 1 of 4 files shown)


till-m:
Hey @phi-friday,

You're right that enums should be used, but the current Logger setup has deeper issues beyond that. The `ScreenLogger`, `JSONLogger`, etc., aren't working well for several reasons. I’d prefer a full overhaul, moving away from the observer pattern entirely, so from my perspective this PR isn’t worth merging as a quick fix.

phi-friday:
i agree.

Issue #370, opened 2022-10-24
Cannot serialize Queue


chengscott:
The following code works with `bayesian-optimization==1.2.0`, but fails with `bayesian-optimization==1.3.0` (or the latest one).

```python
import ray
from ray import tune
from ray.tune.search import ConcurrencyLimiter
from ray.tune.search.bayesopt import BayesOptSearch


def scoreFn(config):
    print(config)
    tune.report(
        ret=config['x'] * 1.2 + config['y'] * 1.2
    )

bo = BayesOptSearch(
    space={
        'x': (0.2, 0.95),
        'y': (0.5, 2.5),
    },
    metric='ret',
    mode='min',
)

tuner = tune.Tuner(
    scoreFn,
    tune_config=tune.TuneConfig(
        num_samples=24,
        search_alg=ConcurrencyLimiter(bo, max_concurrent=4),
    ),
)
results = tuner.fit()
```
The error in `bayesian-optimization==1.3.0`:
```shell
Traceback (most recent call last):
  File "/run/media/adam/Data/search_bo.py", line 38, in <module>
    tuner = tune.Tuner(
  File "/run/media/adam/Data/venv-py310/lib/python3.10/site-packages/ray/tune/tuner.py", line 152, in __init__
    self._local_tuner = TunerInternal(**kwargs)
  File "/run/media/adam/Data/venv-py310/lib/python3.10/site-packages/ray/tune/impl/tuner_internal.py", line 123, in __init__
    pickle.dump(self, fp)
  File "/run/media/adam/Data/venv-py310/lib/python3.10/site-packages/ray/cloudpickle/cloudpickle_fast.py", line 57, in dump
    ).dump(obj)
  File "/run/media/adam/Data/venv-py310/lib/python3.10/site-packages/ray/cloudpickle/cloudpickle_fast.py", line 627, in dump
    return Pickler.dump(self, obj)
TypeError: cannot pickle '_thread.lock' object
```
After bisecting, the code works if I revert [commit 8cc8d9b](https://github.com/fmfn/BayesianOptimization/commit/8cc8d9bffd38cd8c5e7597f7ce92e1c9810d64fc).
A quick fix would be reverting that commit in the master branch.

bwheelz36:
thanks for reporting. will fix ASAP - will also add some new test cases since that commit did pass all existing tests.

bwheelz36:
hi, can you provide instructions for installing the libraries necessary to run this example?
[...]

PR #508, opened 2024-07-26
Fix replace list with deque

phi-friday:
Replace `queue` implemented using `list` with `deque`.

The implementation with `list` appears to be due to the use of `append()` and #370 .
If `append()` is used, `deque` is more efficient, and #370 is also solved.


till-m (approved):
Seems like a good thing to use a built-in for this instead of writing it ourselves!

PR #509, opened 2024-07-26
PR 1/2: Add versioned docs

till-m:
Fix #504 

Demo: https://till-m.github.io/BayesianOptimization/

This adds a version dropdown to our docs page. In the future, any new release will be added to this version dropdown automatically. Similarly, it will keep an updated version of the `master` docs.

I had to switch to `sphinx-immaterial` to get this to work. I’ve also cleaned up the docs a bit and made everything look nicer (imo).

I've also trimmed the README significantly.

## How does it work?
There’s a workflow that runs on push (master), pull request and release. This workflow does a number of things.

### 1) `build-docs-and-publish`
1. It builds the docs (to use but also to check whether they actually build).
2. It determines the current state of the git ref and sets a `target` variable accordingly.
    1. If running on master, set `target` to master
    2. If running on a release, set `target` to release version
    3. If running due to PR, set `target` to empty string
3. If the `target` is not empty string, it deploys the `html` folder of the built docs to `gh-pages/<target>`. It overwrites anything in `<target>` but ignores any other content in `gh-pages`

### 2) `Update docs versions JSON`
Only runs if the `target` is not empty
1. Clones the `gh-pages` branch and reads out the folders within (ignoring anything starting with `.`)
2. Reads out the folder names, determines which are referring to versions and which not.
3. Finds the maximum version
4. Generates a JSON object encoding the present folders/version and the maximum version
5. Pushes that JSON file to the `gh-pages` (ignoring any other files in the branch).

When you open the docs page, the landing page reads out the versions.json and redirects you to `stable`. If there is no version tagged `stable`, it will redirect to `master`.

### On merging this

Requires a sister PR to be merged first, prepping the `gh-pages` branch: #510 

I’ve purged the gh-pages branch and added docs for the `v1.5.1` release. This way I could a) test the `publish` workflow, b) ensure that there is a stable version present, c) make persistent docs for `v.1.5.1` and d) have a more functional demo.

## Credit

Based on [github.com/brechtm/rinohtype](github.com/brechtm/rinohtype)'s workflow.

PR #510, opened 2024-07-26
PR 2/2: Add versioned docs

till-m:
Sister PR to #509 -- please see there for information about this PR.


till-m:
I know this PR looks huge, but it really just sets the `gh-pages` branch up to look like this.
The only file in this PR that could require a review is `index.html`. Everything else is autogenerated.
Screenshot 2024-07-26 at 20 18 33

PR #511, opened 2024-07-27
docs: replace " : " with ": "

phi-friday:
related: #509 (especially https://github.com/bayesian-optimization/BayesianOptimization/pull/509#issuecomment-2253749844)
reproduce: ```find bayes_opt -type f -name "*.py" | xargs sed -i '' 's/ : /: /g'```

It would be nice if this could be enforced by a `ruff` or `pydocstyle` rule, but I haven't found a way.

### DIFF
--- a/bayes_opt/acquisition.py
+++ b/bayes_opt/acquisition.py
@@ -27,7 +27,7 @@ class AcquisitionFunction(abc.ABC):
 
     Parameters
     ----------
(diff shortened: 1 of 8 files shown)


till-m:
Closing -- the rendering problem was due to a missing extension. Let's keep the `param : type` syntax for now, as it's the numpy style.

PR #512, opened 2024-07-28
fix: Incorrectly modified `ScreenLogger`

phi-friday:
Before I applied #497  it worked like this
https://github.com/bayesian-optimization/BayesianOptimization/blob/bc9af23981cb22b98344424b5c2544b741b74938/bayes_opt/logger.py#L167

but after,
https://github.com/bayesian-optimization/BayesianOptimization/blob/16c00f3b7867d7cf43047b2c80d4b6e54156b5e8/bayes_opt/logger.py#L164

I expect it to work the same as before.

### DIFF
--- a/bayes_opt/logger.py
+++ b/bayes_opt/logger.py
@@ -3,9 +3,9 @@
 from __future__ import annotations
 
 import json
-from collections import deque
 from contextlib import suppress
 from pathlib import Path
+from typing import Any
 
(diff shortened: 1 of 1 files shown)


till-m (approved):
Thanks, good fix. I was just wondering about this.

Will merge after the docs change.

till-m:
Thanks!

Issue #493, opened 2024-07-09
support PEP 484


phi-friday:
**Is your feature request related to a problem? Please describe.**

No `type hint` is written, resulting in a static analysis that differs from annotations.
ex: `xi` accepts `float`, but is inferred to be `int` because of the default.

![스크린샷 2024-07-09 오전 9 04 26](https://github.com/bayesian-optimization/BayesianOptimization/assets/19867547/f3458aab-d607-44a7-ad51-74d13836b1c8)

This also causes some issues with static analysis and using the IDE.

**Describe the solution you'd like**
Add the correct `type hint`.

**References or alternative approaches**
https://peps.python.org/pep-0484/

**Additional context**
If you decide to support `type hint`, I'll contribute.
However, since a lot of changes are planned in #447 , i'll start with modules that are not related to #447 .

**Are you able and willing to implement this feature yourself and open a pull request?**
- [X] Yes, I can provide this feature.

phi-friday:
I have one question while writing #494.
`TargetSpace.register()` requires `float` as `constraint_value`, and if `constraint_value` is restricted to `float`.
then the return type of `ConstraintModel` will be fixed to `float`,
and the value of `constraint` in `TargetSpace.max()` will also be fixed to `float`.

However, when I checked `logger.py`, it seems to happen in the case of `np.ndarray`.

Is the comment in `TargetSpace.register()` wrong?

till-m:
In case of multiple constraints this can indeed be an `np.ndarray`. See e.g. [this test](https://github.com/bayesian-optimization/BayesianOptimization/blob/3a5e56c7d0f96affdebfebc318b02f71f0462be3/tests/test_constraint.py#L114).

till-m:
That is to say -- yes, I wrote documentation that's wrong 😬

phi-friday:
> In case of multiple constraints this can indeed be an `np.ndarray`. See e.g. [this test](https://github.com/bayesian-optimization/BayesianOptimization/blob/3a5e56c7d0f96affdebfebc318b02f71f0462be3/tests/test_constraint.py#L114).

Thanks, that makes it a little clearer where to fix it.

By the way, can I continue with this #494 ?

till-m:
> By the way, can I continue with this https://github.com/bayesian-optimization/BayesianOptimization/pull/494 ?

[...]

PR #507, opened 2024-07-25
Fix add typehint without generic

phi-friday:
related: #493 

I didn't use the generic and added type hints using primitive types in most cases.

Therefore, I ignored some types with `Any`.
I also ignored some errors from the static type checker.

Hopefully, this will provide a better IDE experience than before.

PR #520, opened 2024-09-20
Apply PEP561

phi-friday:
closes: #493 
PEP561: https://peps.python.org/pep-0561/

Based on the previous discussion, type hinting using `TypeVar` was ruled out as it would be a barrier to entry for casual users. Instead, #507  was merged in.

In my opinion, this library fully meets the needs of the existing contributors and users, 
so it makes sense to add `py.typed` to let users know that this library provides sufficient type hints.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,10 +5,7 @@ description = "Bayesian Optimization package"
 authors = ["Fernando Nogueira"]
 license = "MIT"
 readme = "README.md"
-packages = [
-    {include = "bayes_opt"},
-    {include = "bayes_opt/py.typed"},
-]
+packages = [{include = "bayes_opt"}]
 
 [tool.poetry.dependencies]
 python = "^3.9"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -5,7 +5,10 @@ description = "Bayesian Optimization package"
 authors = ["Fernando Nogueira"]
 license = "MIT"
 readme = "README.md"
-packages = [{include = "bayes_opt"}]
+packages = [
+    {include = "bayes_opt"},
+    {include = "bayes_opt/py.typed"},
```
till-m, line 10: It seems this explicit include [might not actually be necessary](https://jugmac00.github.io/blog/bite-my-shiny-type-annotated-library/#poetry).

till-m (commented):
How do we check the effect of this change?

pyproject.toml
phi-friday: In local, I ran the `poetry build` command to verify that `py.typed` was included.
But I prefer to be explicit rather than implicit.
Of course, this is just my personal preference, but if you prefer simplicity, it seems better to remove it.
I'll fix it as soon as you point it out.

phi-friday: I fixed it

PR #523, opened 2024-09-20
Fix min window type check

phi-friday:
`minimum_window` should be typechecked using `Mapping`, not `dict`.
Also, `minimum_window_value` should be typechecked using `Sequence`, not `list`.

And I found it while checking it, is it possible for `minimum_window` to be `None`?
The annotation and docstring on `minimum_window` don't match.
Also, there doesn't seem to be any handling for the case where `minimum_window` is `None` in `SequentialDomainReductionTransformer._trim()`.
https://github.com/bayesian-optimization/BayesianOptimization/blob/10ef3be32bd2dd2e9739f430db98a71ae440c4d8/bayes_opt/domain_reduction.py#L211-L212
In my opinion, `typing.Optional` was used to mean `minimum_window` has a default value. https://github.com/bayesian-optimization/BayesianOptimization/blob/a6697aabf590c9c6f3000b55c5195401b37cff98/bayes_opt/domain_reduction.py#L56

So it would be nice to add a process to `__init__` to either not accept `None`, or to convert to the default value when it is `None`.

### DIFF
--- a/bayes_opt/domain_reduction.py
+++ b/bayes_opt/domain_reduction.py
@@ -8,6 +8,7 @@ simple domain reduction scheme for simulation-based optimization"
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
+from collections.abc import Iterable, Mapping, Sequence
 from typing import TYPE_CHECKING, Any
 from warnings import warn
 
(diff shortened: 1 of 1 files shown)


till-m:
Give me some time to look into this :)

till-m:
You're right, this should not accept `None`!

phi-friday:
> You're right, this should not accept `None`!

Should I modify it so that it doesn't accept `None`? Or should I change it to default?

till-m:
it would be great if you remove the `None` option as part of this PR 👍

phi-friday:
> it would be great if you remove the `None` option as part of this PR 👍

I fixed it.

PR #525, opened 2024-09-20
fix: replace acquisition with exception when import error

phi-friday:
Importing errors from `acquisition` does not currently cause any problems at runtime.
However, it is done implicitly and may cause problems in the future.
Therefore, it is safer to explicitly import from `exception`.


till-m:
Thanks for your contribution!

PR #545, opened 2025-02-11
upgrade poetry2.0 & apply pep621

phi-friday:
What's different in poetry2.0: [Announcing Poetry 2.0.0](https://python-poetry.org/blog/announcing-poetry-2.0.0/)

Since poetry2.0 now supports pep621, I was able to modify pyproject.toml to conform to the standard spec.
(I kept the modifications to a minimum).
(Since [install-poetry](https://github.com/snok/install-poetry) doesn't support poetry2.0, I've changed how to install poetry from github action. https://github.com/snok/install-poetry/issues/164)

However, it does not yet support pep735 (dependency groups), so the proprietary specification `tool.poetry.group` is retained.
However, this too will be fixed in the future. https://github.com/python-poetry/poetry/issues/9751

If we use poetry in the future, it seems like a good idea to upgrade poetry to 2.0 if possible, as all improvements will only be for 2.0 or later.

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -26,13 +26,10 @@ jobs:
         uses: olegtarasov/get-tag@v2.1
       - name: Install pandoc
         run: sudo apt-get install -y pandoc
-      - name: Install the latest version of uv
-        uses: astral-sh/setup-uv@v5
-        with:
-          version: "latest"
       - name: Install Poetry
-        run: |
-          uv tool install poetry
+        uses: snok/install-poetry@v1
+        with:
+          version: 'latest'
       - name: Install package and test dependencies
         run: |
           poetry install --with dev,nbtools
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -26,8 +26,13 @@ jobs:
         uses: olegtarasov/get-tag@v2.1
       - name: Install pandoc
         run: sudo apt-get install -y pandoc
+      - name: Install the latest version of uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          version: "latest"
       - name: Install Poetry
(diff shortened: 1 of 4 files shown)


till-m (approved):
LGTM, generally speaking. Will merge soon.

It seems a bit weird to install `uv`, to then install `poetry` later. Using just `uv` would be nice in the future, but I'm a bit worried about breaking the publishing workflow. Something to keep in mind for the future.

phi-friday:
> LGTM, generally speaking. Will merge soon.
> 
> It seems a bit weird to install `uv`, to then install `poetry` later. Using just `uv` would be nice in the future, but I'm a bit worried about breaking the publishing workflow. Something to keep in mind for the future.

I prefer uv to poetry, so I agree with the later change to uv.
(The late support for PEP621 was a big part of what made me move to uv).

But I also agree with the concerns you mentioned.
uv is a good tool, but it will cause problems for other workflows that are already used to poetry and are already aligned with poetry.
Also, using uv as a project package manager will force existing contributors to re-familiarize themselves with uv.
So it's kind of weird, but I've only made some adaptations to uv in github actions.

PS1.
Applying uv is not necessarily just for installing poetry.
I use uv to generate requirements with numpy conditions applied.
This was hard to do cleanly with poetry alone.

PS2.
The reason I changed the way I install poetry is fixed.
https://github.com/snok/install-poetry/issues/164#issuecomment-2653570738
Since python>=3.9 doesn't cause the problem, I think we can keep using install-poetry as is.
(`BayesianOptimization` requires python>=3.9).
I'll fix it soon.

till-m:
> I'll fix it soon.

In that case I will wait with merging this :)

phi-friday:
> > I'll fix it soon.
> 
> In that case I will wait with merging this :)

i fixed it

till-m:
Thanks for the contribution :)

PR #547, opened 2025-02-17
Add logging with state to Bayesian Optimizer

adrianmolzon:
Hey all, long time listener first time contributor. I've created two new methods for the BayesianOptimizer class to allow users to save and load their optimizers while maintaining state. I've added a new section to the basic tour to demonstrate this cleaner method for saving and loading their optimizer.

A couple notes, users using a custom acquisition function will have to write their own get_acquisition_params and set_acquisition_params class functions in order to preserve exact state, otherwise there might be a loss of state when saving and loading. This should still be strictly better than the previous flow for saving and loading optimizers which ignored the acquisition parameters.

If this pull request gets merged into master, I'd like to remove the previous saving methodology, but I'll save that for a future PR.


adrianmolzon:
@till-m

till-m:
@adrianmolzon I realized the coverage bot is broken, so let me fix that first before moving on with this PR

bayes_opt/acquisition.py
```diff
@@ -69,18 +69,57 @@ def __init__(self, random_state: int | RandomState | None = None) -> None:
...
+
+    def set_acquisition_params(self, params: dict) -> None:
+        """Set the acquisition function parameters.
+
+        Parameters
+        ----------
+        params : dict
+            Dictionary containing the acquisition function parameters.
+        """
+        return {}
```
till-m, line 122: While I don't think that we should enforce that these methods are defined for subclasses (i.e. we don't need to make them `abstractmethod`s), we should make them raise a `NotImplementedError` so that if people try to store/load with custom acqs without having custom, properly defined `get`/`set` methods, it will error out instead of letting them believe that it just magically knows which parameters to store and load

bayes_opt/acquisition.py
```diff
@@ -129,6 +168,15 @@ def suggest(
...
 
+    def _fit_gp(self, gp: GaussianProcessRegressor, target_space: TargetSpace) -> None:
+        # Sklearn's GP throws a large number of warnings at times, but
+        # we don't really need to see them here.
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            gp.fit(target_space.params, target_space.target)
+            if target_space.constraint is not None:
+                target_space.constraint.fit(target_space.params, target_space._constraint_values)
+
```
till-m, line 179: I think you moved this function declaration. Any specific reason why?

bayes_opt/acquisition.py
```diff
@@ -1035,3 +1199,42 @@ def suggest(
...
+        dict
+            Dictionary containing the current acquisition function parameters.
+        """
+        return {
+            "base_acquisitions_params": [acq.get_acquisition_params() for acq in self.base_acquisitions],
+            "gains": self.gains.tolist(),
+            "previous_candidates": self.previous_candidates.tolist()
+            if self.previous_candidates is not None
+            else None,
+            "random_states": [acq._serialize_random_state() for acq in self.base_acquisitions]
```
till-m, line 1217: Are the random states for the base_acqs not already stored in Line 1212 under `"base_acquisition_params"`?

examples/optimizer_state.json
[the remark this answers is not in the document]
till-m: The same holds for `logs.log`.

bayes_opt/acquisition.py
adrianmolzon: Each of these base acquisition functions in base_acquisitions have:

1. A random state initialized inside of their acquisition function
2. A high level random state from the BayesianOptimizer

The base_acquisition_params key holds the acquisition function parameters, which will include inside it 1. The random_states key holds the random state of the BayesianOptimizer random state, 2.

examples/optimizer_state.json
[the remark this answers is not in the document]
adrianmolzon: removed

bayes_opt/acquisition.py
```diff
@@ -100,26 +100,42 @@ def _deserialize_random_state(self, state_dict: dict | None) -> None:
...
+    def _fit_gp(self, gp: GaussianProcessRegressor, target_space: TargetSpace) -> None:
+        # Sklearn's GP throws a large number of warnings at times, but
+        # we don't really need to see them here.
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            gp.fit(target_space.params, target_space.target)
+            if target_space.constraint is not None:
+                target_space.constraint.fit(target_space.params, target_space._constraint_values)
+
+    def get_acquisition_params(self):
```
till-m, line 112: I think the return type of this should be `dict[str, Any] | NoReturn`

till-m:
I think this is ready to go, I will merge this in the coming days if there are no objections.

till-m:
Thanks for your contribution!

PR #556, opened 2025-03-25
Remove event based logging pattern

adrianmolzon:
Hey all, I'm back with another pull request! This one removes some of the old functionality for saving logs since I've created a better way of saving state. In doing so, I've tried to simplify the file structure, mostly by deleting unnecessary files, but also by moving all the logging methods to a Logger class in logger.py. This class is now invoked on initialization of the BayesianOptimization object, and exposes the steps previously invoked by the Observable pattern, optimization_start, optimization_step, and optimization_end. I think this unification is cleaner. Changes to the format of the Logger can still be made, they are just stored under optimizer.logger._whateveryouwant_.


till-m:
Hi @adrianmolzon, could you rebase based on upstream? This hopefully also fixes the conflicts :)

Issue #515, opened 2024-08-08
`ScreenLogger.format_number` does not account for minus sign, potentially truncating exponents


till-m:
c.f. #514

**Demo**
```python
import numpy as np
from bayes_opt import BayesianOptimization


x = np.array([
    0.00869858, 0.01304788, 0.02174646, 0.03044504, 0.03914363,
    0.04784221, 0.05654079, 0.06523938, 0.07393796, 0.08263655,
    0.09133513, 0.10003371, 0.1087323 , 0.11743088, 0.12612946,
    0.13482805, 0.14352663, 0.15222522, 0.1609238 , 0.16962238,
    0.17832097, 0.18701955, 0.19571813, 0.20441672, 0.2131153 ,
    0.22181388, 0.23051247, 0.23921105, 0.24790964, 0.25660822,
    0.2653068 , 0.27400539, 0.28270397, 0.29140255, 0.30010114,
    0.30879972, 0.31749831, 0.32619689, 0.33489547, 0.34359406,
    0.35229264, 0.36099122, 0.36968981, 0.37838839, 0.38708698,
    0.39578556, 0.40448414, 0.41318273, 0.42188131, 0.43057989
])
y = np.array([
    54688.54691276, 56090.81734642, 49640.37335158, 54688.54691276,
    50481.73561177, 59660.23299573, 54364.94604345, 54408.09282602,
    50729.19510007, 52695.84682282, 55089.19560809, 52859.49852102,
    49472.10089954, 43470.38344347, 40907.61334058, 41525.29864839,
    37478.86431783, 34255.46345085, 37406.51129791, 34948.89388507,
    33449.28010048, 29545.51192782, 29728.1331936 , 31327.3181988 ,
    26356.96060002, 28045.40867321, 28733.31492368, 23405.1683291 ,
    26052.70858327, 23672.22630383, 22620.23125773, 25574.74171866,
    25370.30815361, 23985.10323843, 22192.45381967, 22989.33499691,
    23396.7861397 , 20921.87487021, 20269.18172291, 22365.32590395,
    20722.44085298, 19614.89124674, 22271.35394637, 18471.28640201,
    21553.99947469, 20063.25389699, 19631.78607125, 21609.72541978,
    20773.84137289, 19674.27911469
])

rt_max = 2*x[-1]
fd0 = 19896.45379


def distance(cd, rc, rt, fd):
    if rc > rt:
        return -1e10
    kdens = cd * (
[...]

PR #557, opened 2025-03-26
Logging

adrianmolzon:
Hey all, I'm back with another pull request! This one removes some of the old functionality for saving logs since I've created a better way of saving state. In doing so, I've tried to simplify the file structure, mostly by deleting unnecessary files, but also by moving all the logging methods to a Logger class in logger.py. This class is now invoked on initialization of the BayesianOptimization object, and exposes the steps previously invoked by the Observable pattern, optimization_start, optimization_step, and optimization_end. I think this unification is cleaner. Changes to the format of the Logger can still be made, they are just stored under optimizer.logger.whateveryouwant.


bayes_opt/bayesian_optimization.py
```diff
@@ -105,6 +76,27 @@ class BayesianOptimization(Observable):
         This behavior may be desired in high noise situations where repeatedly probing
         the same point will give different answers. In other situations, the acquisition
         may occasionally generate a duplicate point.
+
+    Attributes
```
till-m, line 80: @adrianmolzon could you build the docs and check whether this renders correctly?

bayes_opt/logger.py
```diff
@@ -62,11 +33,21 @@ class ScreenLogger(_Tracker):
...
-    def __init__(self, verbose: int = 2, is_constrained: bool = False) -> None:
+    def __init__(
+        self,
+        verbose: int = 2,
+        is_constrained: bool = False,
+        params_config: Mapping[str, ParameterConfig] | None = None,
+    ) -> None:
         self._verbose = verbose
         self._is_constrained = is_constrained
+        self._params_config = params_config
```
till-m, line 44: It's not ideal that we need to copy this object over whole, but maybe there is no other way (?)

bayes_opt/logger.py
```diff
@@ -200,105 +203,97 @@ def _header(self, instance: BayesianOptimization) -> str:
...
-    The resulting file can be used to restart the optimization from an earlier state.
+        if self._previous_max is None or current_max["target"] > self._previous_max:
+            self._previous_max = current_max["target"]
+            self._previous_max_params = current_max["params"]
 
-    Parameters
-    ----------
-    path : str or os.PathLike
-        Path to the file to write to.
+    def _time_metrics(self) -> tuple[str, float, float]:
```
till-m, line 244: What is this used for?

bayes_opt/logger.py
```diff
@@ -200,105 +203,97 @@ def _header(self, instance: BayesianOptimization) -> str:
...
+            The result dictionary for the most recent step.
 
-        self._update_tracker(event, instance)
+        current_max : dict[str, Any] | None
+            The current maximum target value and its parameters.
+        """
+        is_new_max = self._is_new_max(current_max)
+        self._update_tracker(current_max)
+
+        if self._verbose != 1 or is_new_max:
```
till-m, line 289: Why the verbose check here 🤔

bayes_opt/logger.py
adrianmolzon: Good question. Right now, nothing. In the old logging functionality, information about the time and the time elapsed was saved to the log file. Do you still want that information? I would have to decide where to save that information and when to present it, since that log file doesn't exist anymore.

adrianmolzon: I'm happy to remove this function and the associated tests if it's not necessary information

bayes_opt/logger.py
adrianmolzon: Because the old system had the same functionality, and I mindlessly copied the code to the new system. I think the idea makes sense, we should only print if we have a new max or if we want more verbosity, although the check for not equals 1 is definitely weird, arguably it should be >= 1?

bayes_opt/logger.py
till-m: yeah, let's delete that :)

bayes_opt/logger.py
till-m: Writing this out for my sanity:
- `verbose=0`: No printing to console
- `verbose=1`: Printing maxima only
- `verbose=2`: Print all steps

Maybe this handles the colouring, and in the `verbose=1` case we don't want any colour as the whole table would then be highlighted, making that kinda pointless?

In any case, could maybe verify the behaviour is what one would expect?

adrianmolzon: Yeah we have the same expected logic, I'll make sure we're good

adrianmolzon:
> @adrianmolzon if you feel like losing your sanity you could also have a look at tackling #515.

Yeah this sucked but I've made some changes to fix this. To make things easier, any number over a million will be converted into scientific notation, hope this doesn't ruin anybody's day.

till-m:
I think this is good to go! Will merge after easter :)

till-m:
thanks! :)

PR #561, opened 2025-04-27
Stop duplicating ensure_rng in AcquisitionFunction class

Gerrit0:
I noticed this while porting the project to another environment


till-m:
Thanks for the PR! Out of curiosity, what are you porting this project to, and why? :)

Gerrit0:
I needed an implementation of it in C++ - didn't want to have to call Python in order to optimize things. It was super easy to get up and running with the library with a Python plugin, so it made sense to me to port the parts I needed to C++ rather than try and find some library in that language with the same API

PR #562, opened 2025-04-27
Fix weird formatting

Gerrit0:
Cleans up another little weird bit of code I noticed


till-m:
@Gerrit0 this is indeed weird code, but it looks like your change tripped up the linter. Could you check back and fix that? :)

Gerrit0:
@till-m fixed!

PR #566, opened 2025-05-22
polish code/spring clean

till-m:
This PR contains various improvements:
- Makes the `random_state` not a property of the acquisition function, but something to be provided during `.suggest`
    - Before, when choosing a non-default acquisition function and wanting deterministic behaviour, one had to set a lot of states manually. Now, the optimizer state will be used for the acquisition function
- Rename `n_lbfgs_b` to `n_smart` since we now have the DE optimization
- fix another bug related to acq maximization, which caused seeding of the smart optimization to use suboptimal points
    - this also caused a test to pass that should've failed. I fixed this test.


till-m:
@coderabbit review

Issue #499, opened 2024-07-15
Support high-dimensional BO


jacktang:
**Is your feature request related to a problem? Please describe.**
Bayesian optimization for high-dimensional constraint problem

**Describe the solution you'd like**
The solution called Sparse Axis-Aligned Subspace BO (SAASBO) is described in the paper: https://arxiv.org/abs/2103.00349

**References or alternative approaches**
* https://arxiv.org/abs/2103.00349

**Are you able and willing to implement this feature yourself and open a pull request?**
- [x] Yes, I would like to try to implement it

till-m:
Hi @jacktang,

I haven't had the time to read this paper, but considering that this is essentially an acquisition strategy, do you think you could implement this simply by [adding a new acquisition function](https://github.com/bayesian-optimization/BayesianOptimization/blob/master/examples/acquisition_functions.ipynb)?

jacktang:
Hello @till-m 

Yes,  I am going to implement it the way you suggested. And I noticed that you had merge the acquisition function branch to master. Great job! 👍

MandaKausthubh:
Hi @jacktang,

Can we use a Python Library, such as numpyro, for this, which simplifies and provides a pre-packaged version of the No U-Turn Sampling required in SAASBO?

jacktang:
Hello @MandaKausthubh, actually I used numpyro for inner saasgp class. now i am stuck in  solving the high dimensional constrained problem, and plan to submit the code in next two or three weeks.

jacktang:
Hello @till-m, SAASBO may bring some changes : 1) sample based GP (SAASGP) rather than analytical GP (standard GP from sklearn) 2) Sample-based acquisition functions.

jacktang:
I organized the HDBO implementation which mentioned in paper "Standard Gaussian Process is All You Need for High-Dimensional Bayesian Optimization" [here](https://github.com/jacktang/HDBO/blob/main/hdbo-torch.py). Check it if you are interested in it. And try to implement the idea by sklearn GP in next weeks.

jacktang:
Hello @till-m , I had interests to learn the paper, and now only implemented the recommended HDBO by gpytorch. It's great you've draft and ready to push :). I'd like to test the implementation especially for the constrained HDBO if needed.

jacktang:
[...]

PR #569, opened 2025-05-31
Added basic SAASBO implementation

MandaKausthubh:
This PR also adds Pyro as a requirement, which is used for probabilistic sampling with MCMC and NUT Sampler. This adds SAASBO as a child class for the `BayesianOptimization` class. I need feedback for testing as the SAASBO class requires a large number of variables.


till-m:
closed for now, see: #499 for the discussion on HDBO

PR #577, opened 2025-06-24
Add parameter validation

spline2hg:
Added parameter validation for acquisition functions

## Summary by CodeRabbit

- **New Features**
  - Added support for custom acquisition functions in Bayesian optimization setup.
- **Bug Fixes**
  - Improved input validation for acquisition function parameters to ensure values are within valid ranges, preventing misconfiguration and enhancing reliability.

### DIFF
--- a/bayes_opt/acquisition.py
+++ b/bayes_opt/acquisition.py
@@ -612,8 +612,8 @@ class ProbabilityOfImprovement(AcquisitionFunction):
         exploration_decay_delay: int | None = None,
         random_state: int | RandomState | None = None,
     ) -> None:
-        if xi <= 0:
-            error_msg = "xi must be greater than 0."
+        if xi < 0:
+            error_msg = "xi must be greater than equal to 0."
             raise ValueError(error_msg)
         if exploration_decay is not None and not (0 < exploration_decay <= 1):
             error_msg = "exploration_decay must be greater than 0 and less than or equal to 1."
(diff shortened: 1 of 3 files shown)


spline2hg:
@till-m Just to confirm — xi should be positive, right? I noticed that in the tests we're creating acquisition function instances with xi=0. Since we're adding validation, should we update those test cases accordingly?

till-m:
I did some quick googling (and thinking) and I think `xi=0` is possibly reasonable, whereas negative `xi` is theoretically possible but probably never useful. For now I propose we allow anything non-negative :)

Sidenote: I realize that one of the `decay_delay`'s has lost it's docstring (presumably that is my fault): [link](https://github.com/bayesian-optimization/BayesianOptimization/blame/8e7a2862c051dfe0d9d3ad5bebcefba92c387833/bayes_opt/acquisition.py#L768). Would you mind also fixing that in this PR?

spline2hg:
@till-m For the type annotations, would you recommend doing it like this?

```python
kappa: Annotated[float, "non-negative"] = 2.576
exploration_decay: Annotated[float, "range (0, 1]"] | None = None
exploration_decay_delay: Annotated[int, "non-negative"] | None = None
```

Or should we define reusable types in a separate `typing` module, like:

```python
NonNegativeFloat = Annotated[float, "non-negative"]
```
we can also use annotated_types for Ge and Gt , what do u recommend ?

spline2hg:
@till-m Just to clarify — are you saying that we can keep the types as they are, like:

```python
kappa: float = 2.576
exploration_decay: float | None = None
exploration_decay_delay: int | None = None
```
and not necessarily use any `NonNegativeFloat` or `UnitInterval`?

till-m:
yes, exactly

spline2hg:
okay so is there anything more i can add in this pr?

PR #584, opened 2025-07-12
Update pyproject.toml to use PEP-621 dependency specification

jameskrach:
- Updates pyproject.toml to use the PEP-621 specification format for optional dependencies
- Commits a known working set of dependencies for reproducibility across developers


pyproject.toml
```diff
@@ -25,39 +25,30 @@ dependencies = [
...
+    "nbformat>=5.9.2,<6.0",
+    "nbconvert>=7.14.2,<8.0",
+    "jupyter>=1.0.0,<2.0",
+    "matplotlib>=3.0,<4.0",
+    "nbsphinx>=0.9.4,<1.0",
+    "sphinx-immaterial>=0.12.0,<1.0",
+    "sphinx>=7.0.0,<8.0; python_version<'3.10'",
+    "sphinx>=8.0.0,<9.0; python_version>='3.10'",
+    "sphinx-autodoc-typehints>=2.3.0,<3.0; python_version<'3.10'",
+    "sphinx-autodoc-typehints>=2.4.0,<3.0; python_version>='3.10'",
```
till-m, line 46: Let's drop having two dependency groups which are both effectively `dev`. Just merge them together as `dev` group.

requirements-dev.txt
till-m, line 1: This file somewhat confuses me. Why do we need this over the dependency specification in `pyproject.toml`?

till-m (changes requested):
Hey @jameskrach,

thanks for taking charge here! As already mentioned elsewhere, I think cleaning up the project file is necessary!
Please have a look at the test failure -- this may already be fixed by merging the dependency groups.

till-m:
FYI for anyone who hasn't contributed yet (i.e. some commit on master) I need to manually approve workflows, but I'll keep an eye on this PR :)

requirements-dev.txt
jameskrach: I was trying to add something to the effect of a lockfile for both the dependencies that a user may have or a developer may have. The files are essentially:
  - `pip install -e . && pip freeze --exclude-editable > requirements.txt`
  - `pip install -e .[dev,nbtools] && pip freeze --exclude-editable > requirements-dev.txt`

Both will fully resolve transitive dependencies in addition to the things explicitly specified in `pyproject.toml`, however if you prefer `uv` as mentioned in the other issue, it may be better to just commit `uv.lock` as it will have both.

till-m: I see where you're coming from, but I don't fully agree with this approach. Since we're developing a library intended to be integrated into a larger codebase, we shouldn't expect end-users to rely on a tightly pinned set of dependency versions. Things should just work with any combination of packages that are valid according to the `pyproject.toml`.

till-m:
Feel free to drop the carets :)

Moreover, I think the rabbit might be right and `ruff` is too tightly version-constrained (and should probably be higher, the current version is seemingly `0.12.3`). Could you also look into that?

jameskrach:
Updated the upper bound constraints and the workflow. Didn't have a chance to change the poetry thing to uv, but that may also involve changing some other things. I can look at it either later or in a follow-up PR if you like.

.github/workflows/python-publish.yml
```diff
@@ -14,10 +14,10 @@ jobs:
...
-        pypi_token: ${{ secrets.PYPI_API_TOKEN }}
-        # python_version: "3.10"
-        # poetry_version: "==1.8" # can lock versions if we want
+    - uses: actions/checkout@v4
+    - uses: astral-sh/setup-uv@v6
+    - run: uv build
+    - name: Publish to pypi
+      env:
+        PYPI_API_TOKEN: ${{ secrets.PYPI_API_TOKEN }}
+      run: uv publish --token "$PYPI_API_TOKEN"
```
till-m, line 23: I've never published anything using `uv`, but looking at this and comparing with the [docs](https://docs.astral.sh/uv/reference/cli/#uv-publish): Do we not need to specify the `index` to publish to?

till-m (approved):
Generally LGTM! Only slightly worried about the publishing workflow as it's hard to test 😬

.github/workflows/python-publish.yml
till-m: Ah you're right! Thanks

till-m:
Thanks for this great PR @jameskrach! I will merge it on friday :)

jameskrach:
Great! Thanks for being open to the contribution!

jameskrach:
@till-m hey - any idea when you’ll create a release with this and the new serialization changes (#582)?

till-m:
Usually I like to wait a bit longer with making releases, but if it matters to you I can do it now. Just let me know.

jameskrach:
It would help me close out an issue on a project that uses this package, but it could wait a week or so if needed.

till-m:
should be up on pypi now and on conda probably by tomorrow :)

jameskrach:
Amazing - thank you again for all your help!

PR #588, opened 2025-08-11
Fixed invalid JSON format from the logs

Fdms-3741:
The previous method saved data in an invalid JSON format that made it incompatible with any JSON parser. This modification proposes a change in the save and load functionality that allows it to write and read valid JSON. 

**Warning**: This will break imports of previously saved data. The data itself will not be altered; thus, manual modification is required to turn the old log files into valid JSON files.

## Summary by CodeRabbit

- New Features
  - Added a utility to load observations from one or multiple JSON log files into an existing optimizer. Supports file paths and iterables, skips duplicates when configured, respects constraints, and returns the updated optimizer.

- Documentation
  - Updated logging function documentation with clearer formatting.


till-m:
Hey @Fdms-3741,

thanks for opening a PR, contributions are very welcome in this repository.

That being said, I'm a bit confused. Is this LLM-generated code? I'm reasonably sure that the state-saving produces valid JSON, however, even if not, this would not be the right way to fix the problem.

till-m:
Hey @Fdms-3741,

I think maybe you're on an older version, the save/load functionality was changed in #547 and #557. The new one produces correct JSON as far as I can tell. Could you have a look and check what version you're on?

PR #590, opened 2025-08-28
correct the formatting for positive numbers with negative exponents and add additional test

adrianmolzon:
Fixing https://github.com/bayesian-optimization/BayesianOptimization/issues/589#issuecomment-3227851583
I was checking whether a minus sign was present, the formatter will now check that the number is negative when formatting for a negative number

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved numeric sign handling and width-based truncation in log number formatting so scientific notation and negative values display correctly (exponent-aware truncation, sign determined by numeric value).

- **Tests**
  - Added/expanded tests covering small-magnitude scientific notation for positive and negative values to prevent regressions.

### DIFF
--- a/bayes_opt/logger.py
+++ b/bayes_opt/logger.py
@@ -96,9 +96,6 @@ class ScreenLogger:
                 width -= dot_pos
                 if width > 0:
                     result += s[dot_pos : dot_pos + width]
-            else:
-                head = s[:e_pos] if "e" in s else s
-                result += head[:width]
             if "e" in s:
                 result += end
             result = result.ljust(self._default_cell_size)


adrianmolzon:
@LiemuR Thanks again for pointing out this mistake. To be clear, this was just a formatting issue, but the formatting issue has now been fixed.

@till-m Can I get an approval here

till-m (approved):
I'd be fine with merging this as-is, but maybe check if the clanker had some nitpicks worth considering.

till-m (commented):
LGTM, but I think it makes sense to add cover this little bit of added code.

PR #591, opened 2025-09-27
add alternative termination criteria

cballam:
This is a WIP attempt to satisfy some of the goals in this issue thread about adding alternative (not just iteration based) termination criteria for the optimizer: https://github.com/bayesian-optimization/BayesianOptimization/issues/381

New termination criteria are:

- value of the user-defined function
- time taken to run the full optimization process
- absolute tolerance of the improvement in function value

I've added tests to make sure these are functional. The "time taken" one is challenging as this depends on the user's machine (the old "it works on my machine" excuse), but is empirically accurate to within about 200ms for low complexity target/acquisition functions 

Open to any criticism/suggestions here, I am sure the code could be improved but am not sure on the best way to go about it.

### DIFF
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -12,6 +12,8 @@ from os import PathLike
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 from warnings import warn
+from datetime import timedelta, datetime, timezone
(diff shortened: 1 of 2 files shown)


t-muser:
Closing for now, though I would be very happy to reopen, just let me know :)

PR #603, opened 2026-03-26
Remove stale ndarray sorting warnings

ReinerBRO:
Remove the old ndarray sorting warning now that pbounds order is the canonical order. Add regression tests for register/probe with ndarray inputs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed unnecessary warnings when registering or probing with NumPy arrays.

* **Tests**
  * Added tests verifying NumPy array inputs work correctly and parameters properly map to bounds order.

### DIFF
--- a/bayes_opt/bayesian_optimization.py
+++ b/bayes_opt/bayesian_optimization.py
@@ -12,7 +12,6 @@ from collections.abc import Iterable
 from os import PathLike
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
-from warnings import warn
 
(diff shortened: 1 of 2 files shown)


t-muser:
Hey, thanks for the PR!

The warning is not stale (as in outdated), it alerts users that the behaviour has changed from older versions. However, I think it may still be time to remove it, it's been like this for ~9 months now.

Will merge within a few days, unless there are objections :)

t-muser:
Thanks for the contribution!

PR #613, opened 2026-06-05
fix: gphedge duplicates

phi-friday:
While preparing the Python 3.14t support PR, I found a case where `GPHedge` can select an already registered boundary point with the newer numerical dependency stack.

 `GPHedge` selects one candidate from multiple base acquisition functions using a softmax over gains, but the previous implementation did not check whether the selected candidate was already registered. As a result, it could return a duplicate candidate even when another base acquisition had produced a fresh one.

 This did not show up in the existing test environment because the short seeded loop did not select a duplicate candidate there. However, GP/acquisition optimization results can vary across dependency versions, and the newer stack exposed this latent bug.

 When duplicate points are disabled, such a candidate raises `NotUniqueError` if passed directly to `register()`. Even through the public `probe()` path, it wastes an optimization iteration because no new observation is added.

Without this PR, supporting Python 3.14t requires a test-only change to `test_gphedge_integration`, because the newer dependency stack selects a duplicate `GPHedge` suggestion in that test.

## Summary by CodeRabbit

* **Bug Fixes**
  * GPHedge now properly skips duplicate candidates when duplicates are forbidden, resampling from unique candidates to avoid returning previously registered points.

* **Tests**
  * Added test case validating duplicate candidate handling in GPHedge acquisition function.

### DIFF
--- a/bayes_opt/acquisition.py
+++ b/bayes_opt/acquisition.py
@@ -1316,6 +1316,15 @@ class GPHedge(AcquisitionFunction):
         ]
         self.previous_candidates = np.array(x_max)
         idx = self._sample_idx_from_softmax_gains(random_state=random_state)
+        if not target_space._allow_duplicate_points and x_max[idx] in target_space:
+            # If every candidate is a duplicate, keep the original choice.
+            # Avoiding duplicates then requires generating new candidates, which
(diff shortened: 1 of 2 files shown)


t-muser:
nice catch, thanks!

